PR #6, opened 2016-12-23
Observable documentation

dustinlessard-wf:
### Issue

It would be super helpful to see a short comment about what each operator does in our IDEs.

### Solution

Added comments to each operator.  (mostly pulled from rxjs)

rxdart_comments

todo: There is one operator that I could use some help on : tween

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -10,5 +10,5 @@ dev_dependencies:
   browser: ^0.10.0+2
   faker: ^0.0.4
   react: ^0.8.5
-  test: ^0.12.5+1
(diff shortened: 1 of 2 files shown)


dustinlessard-wf:
I'm guessing the CI failure is related to https://github.com/dart-lang/test/issues/515 possibly?

pubspec.yaml
```diff
@@ -10,5 +10,5 @@ dev_dependencies:
   browser: ^0.10.0+2
   faker: ^0.0.4
   react: ^0.8.5
-  test: ^0.12.5+1
```
frankpepermans: Tests were failing indeed with the latest test version, this is fixed meanwhile and pushed to the main branch, can you remove this dependency change?

PR #12, opened 2017-01-08
Add `just` constructor

brianegan:
I've found myself doing `new Observable.fromIterable(<int>[1])` a few times now, and realized I miss the `new Observable.just(1)` constructor from other Rx implementations. I know it's a simple change, but one that might be nice to have :)

Would you be open to this PR which adds the `just` constructor from the ReactiveX spec? http://reactivex.io/documentation/operators/just.html

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ allowing usage of this library in server side (or Flutter) projects as well.
     .sample
     .scan
     .startWith
+    .startWithMany
     .takeUntil
     .timeInterval
     .tap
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -30,6 +30,7 @@ allowing usage of this library in server side (or Flutter) projects as well.
     .sample
     .scan
     .startWith
+    .startWithMany
```
brianegan: Just threw this in there because I saw it was missing

brianegan:
Sure thing! Just pushed a test up. The factory doesn't contain much logic itself as it defers that to other methods, so lemme know if that's enough coverage or if there's more you'd like tested.

PR #31, opened 2017-01-30
Add onErrorResumeNext and onErrorReturn operators

brianegan:
Another couple operators I've used in RxJava for Android apps, and therefore feel like it's good to include them now!

### DIFF
--- a/lib/src/transformers/on_error_resume_next.dart
+++ b/lib/src/transformers/on_error_resume_next.dart
@@ -1,0 +1,51 @@
+import 'dart:async';
+
+StreamTransformer<T, T> onErrorResumeNextTransformer<T>(
+    Stream<T> recoveryStream) {
+  StreamController<T> controller;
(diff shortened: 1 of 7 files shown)


lib/src/transformers/on_error_resume_next.dart
```diff
@@ -0,0 +1,49 @@
...
+
+    controller = new StreamController<T>(
+        sync: true,
+        onListen: () {
+          inputSubscription = input.listen((T data) {
+            controller.add(data);
+          }, onError: (dynamic e, dynamic s) {
+            shouldCloseController = false;
+
+            recoverySubscription = recoveryStream.listen(controller.add,
```
frankpepermans: Should we close the controller when recoveryStream triggers onDone?

brianegan:
Nice catch! Just pushed up a fix :)

PR #36, opened 2017-02-01
Add all the do* operators

brianegan:
Not ready yet -- opening early for discussion. :)

--------

I looked at other implementations and they offer a series of do* operators (doOnSubscribe, doOnUnSubscribe, doOnNext, doOnError, etc.. I went ahead and rounded up all the possible Dart-specific versions of these events, and created all the associated methods.

We ended up with 8 methods in total, all implemented with 1 transformer. 

Questions:

  1. What do you think of the api? I don't mind it right now, but might be harder to change in the future, so perhaps best to ponder a bit now :) We can keep each method separate, or play with some alternatives:
    - `tap(onData: ..., onCancel: ...)`
    - `doOn(data: ..., cancel...)`
    - `act(onData..., onCancel...)`
    - Keep each method as a standalone. There are (probably unimportant) performance tradeoffs here -- using a single method, you could apply all the do* operations to a single transformer, rather than requiring a new transformer for each do* operation. On the other hand, we could implement each transformer individually as well, which would make each one individually faster.
  2. Should we make each transformer responsible for only 1 event? e.g. `doOnDataTransformer` instead of `doTransformer` with 8 optional params?


frankpepermans:
I think having `doOn` is more inline with the current Dart Stream implementation,
since `doOn` is comparable to what `listen` does, and `listen` also provides optional handlers like `onData, onError, onCancel`

For me, the current `tap` implementation can be merged into `doOn`, where we could provide `doOn` with the following signature:

`doOn<T>([void onData(T data)], {void onError(e, s), void onCancel(), ...});`

This way, you would write `tap(something);` as `doOn(something);`
and to notify onCancel: `doOn(onCancel: somethingElse);`

I wouldn't worry too much about performance here, I think the difference would be negligible.

What do you think?

brianegan:
Hey, thanks for the thoughtful response! I'm totally happy to go with `doOn` -- I actually prefer `act(onData: ...)`, but I don't think it's standard Dart or Rx terminology.

Unfortunately, I tried that signature, and Dart can't support both optional-positional arguments and named parameters :( 

Therefore, because I don't think we should enforce setting `onData` the way `listen` does, as I've often just used `doOnError ` or `doOnComplete` by themselves, I'd suggest a method signature like so:

```dart
doOn({void onCancel() = defaultOnCancel,
    void onData(T event) = defaultOnData,
    void onDone() = defaultOnDone,
    void onEach(Notification<T> notification) = defaultOnEach,
    Function onError = defaultOnError,
    void onListen() = defaultOnListen,
    void onPause([Future<dynamic> resumeSignal]) = defaultOnPause,
    void onResume() = defaultOnResume})
```

frankpepermans:
Argh too bad :)

+1 for adding `onDone` as described then!

What about this use case, `stream.doOn()`?
Just do nothing I suppose?

brianegan:
Ah, good catch, that's an interesting case! Yep, it would do nothing in that instance.

To aid users, we can also throw an `assert(onDone != null || onError != null || ...)` check to throw an error when no params are provided as a hint?

dotdotcommadot:
For me, it seems counter-intuitive to write: 
`doOn(onData: (data) => print(data))`

But it would also be strange to write:
`doOn(data: (myData) => print(myData))`
Because 'data' sounds more like a property than a method, and now I have to name my value 'myData' instead of 'data' to keep it clear... I could also use 'value' instead of 'myData', but my point would still remain standing

So I think the most intuitive thing would be:
`act(onData: (data) => print(data))`

brianegan:
>For me, it seems counter-intuitive to write:
> `doOn(onData: (data) => print(data))`
>
>But it would also be strange to write:
> `doOn(data: (myData) => print(myData))`

Totally agree! That's why I preferred `act` as well... Ok, let's break the (not great) Rx `do` convention and go our own way for this one :)

frankpepermans:
I would just do nothing if `act()` occurs, there's no transformation going here, hopefully in a future Dart version we could provide some kind of way to enforce at least one handler here?

brianegan:
Sounds good, I'll go with that for now! What do you think about adding a simple `assert` statement? That would allow us to check for at least 1 method being present, and alert the developer with static messaging that it's empty?

Thanks for helping me figure out these details :)

frankpepermans:
Sounds good 👍

brianegan:
Hey hey :) I ended up putting this together a bit more -- adding tests, docs, etc -- and was still not quite satisfied with `act`. Then it struck me: Maybe we could use `call`? It's idiomatic in Dart, and (I think!?) conveys what's happening.

What do you think? Happy to change it back to `act`, but thought this was an interesting alternative. Maybe also `run`. Anyhow, all other pieces we discussed are implemented :)

```dart
new Observable.just(1).call(onData: (i) => print(i));

 Observable<T> call(
          {void onCancel(),
          void onData(T event),
          void onDone(),
          void onEach(Notification<T> notification),
          Function onError,
          void onListen(),
          void onPause(Future<dynamic> resumeSignal),
          void onResume()})
```

PR #37, opened 2017-02-02
Fail build when formatting is incorrect

brianegan:
This addition will cause the Travis build to fail if the PR contains dart files that aren't properly formatted.

Kind of an annoying one haha, but I keep committing unformatted code, and figure it can't hurt to verify that we're keeping our code properly formatted.

It also fixes all the files that were causing failures :)


frankpepermans:
Good idea, I tend to skip formatting as well, this test will keep us in check :)

PR #43, opened 2017-02-03
added count operator

frankpepermans:
count simply returns an `int` value representing the amount of events emitted by the `Observable`.

You can optionally provide a `predicate` to filter results,
for example say you have an `Observable` emitting values randomly between 0 and 10,
if you only need to keep track of how many items are above 9, you can pass the following `predicate`:

`stream.count((double value) => value > 9.0);`

If you pass no predicate, every event will be counted, for example

```
new Observable.fromIterable(['a', 'b', 'c'])
    .count()
    .listen((int count) => print('count: $count'));
/// 1
/// 2
/// 3
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ var myObservable = Observable.combineLatest3(
 ##### Available Methods
 - bufferWithCount
 - call
+- count
 - debounce  
 - flatMapLatest  
 - flatMap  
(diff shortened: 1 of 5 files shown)


frankpepermans:
Not sure why codecov and Travis are complaining,
might me because codecov seems to incorrectly assume the coverage percentage drops?

event though its own report shows an increase

frankpepermans:
Stream already has length! doh!

PR #44, opened 2017-02-04
Add `range` factory

brianegan:
The sickest range function that ever was. A bit different signature than the RxJava implementation, but it's more flexible as it allows for ranges in both directions.


brianegan:
Hm, I'm actually not sure what's failing on the build :( Everything exits with 0 status, it even runs a job called "after_success" haha. Can you see what's wrong?

frankpepermans:
Happened to me as well, no idea :)

I'll check later today or tomorrow!

PR #53, opened 2017-02-07
added using operator

frankpepermans:
See: http://reactivex.io/documentation/operators/using.html

in rxdart, we can call dispose when subscription.cancel is invoked.

Since a broadcast stream is unaware when all of its subscribers close, I've overridden the asBroadcastStream method to throw instead


brianegan:
Hrm, I've never used this operator and I'm baffled as to how it's useful, but I think this implementation might be just a bit off. The RxJava docs look like this:

```
You pass the using operator three parameters:

  * a factory function that creates a disposable resource
  * a factory function that creates an Observable
  * a function that disposes of the resource

When an observer subscribes to the Observable returned from using, using will use the Observable factory function to create the Observable the observer will observe, while at the same time using the resource factory function to create whichever resource you have designed it to make. When the observer unsubscribes from the Observable, or when the Observable terminates (normally or with an error), using will call the third function to dispose of the resource it created.
```

It looks like this implementation is a bit different -- rather than taking only the DisposableResource, I think we also need to accept a function as the second argument that will create the Stream.

Whatcha think?

PR #54, opened 2017-02-07
Updates the Konami Code Example

dotdotcommadot:
The Konami Code example could use a little cleanup.

### DIFF
--- a/web/konami/konamicode.dart
+++ b/web/konami/konamicode.dart
@@ -3,6 +3,8 @@ import 'dart:html';
 import 'package:collection/collection.dart';
 import 'package:rxdart/rxdart.dart';
 
+// Side note: To maintain readability, this example was not formatted using dart_fmt.
+
 void main() {
   const konamiKeyCodes = const <int>[
     KeyCode.UP,
--- a/web/konami/konamicode.dart
+++ b/web/konami/konamicode.dart
@@ -5,15 +5,19 @@ import 'package:rxdart/rxdart.dart';
 
 void main() {
 
-  const up = 38;
-  const down = 40;
-  const left = 37;
-  const right = 39;
-  const a = 65;
-  const b = 66;
+  const konamiKeyCodes = const <int>[
+    KeyCode.UP,
+    KeyCode.UP,
+    KeyCode.DOWN,
+    KeyCode.DOWN,
+    KeyCode.LEFT,
+    KeyCode.RIGHT,
+    KeyCode.LEFT,
+    KeyCode.RIGHT,
+    KeyCode.B,
+    KeyCode.A
+  ];
 
-  /// The Konami Code: UP, UP, DOWN, DOWN, LEFT, RIGHT, LEFT, RIGHT, B, A
-  const konamiKeyCodes = const <int>[up, up, down, down, left, right, left, right, b, a];
   final result = querySelector('#result');
 
   /// Wrap the `document.onKeyUp` stream so we can use the methods on `Observable
--- a/web/konami/konamicode.dart
+++ b/web/konami/konamicode.dart
@@ -20,7 +20,7 @@ void main() {
 
   final result = querySelector('#result');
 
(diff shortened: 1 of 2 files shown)


web/konami/konamicode.dart
```diff
@@ -1,39 +1,28 @@
...
-  for (int i = 0; i < 10; i++) if (a[i] != b[i]) return false;
-  return true;
-}
+  keyboardInput$
+    // Use map() to get the keyCode
+    .map((event) => event.keyCode)
+    // Use bufferWithCount() to remember the last 10 keyCodes
+    .bufferWithCount(10, 1)
+    // Use where() to check for matching values
+    .where((list) => const IterableEquality<int>().equals(list, codes))
```
brianegan: To really spell out what's going on here, perhaps we can rename the parameter from `list` to `lastTenKeyCodes`? Just thinking it might be easier for newcomers to understand something like `...equals(lastTenKeyCodes, konamiKeyCodes)`

web/konami/konamicode.dart
```diff
@@ -1,39 +1,28 @@
...
-  ];
-  Element result = querySelector('#result');
 
-  // ignore: close_sinks
-  StreamController<KeyboardEvent> controller =
-      new StreamController<KeyboardEvent>();
-  Observable<KeyboardEvent> stream =
-      new Observable<KeyboardEvent>(controller.stream);
+  // The Konami Code: UP, UP, DOWN, DOWN, LEFT, RIGHT, LEFT, RIGHT, B, A
+  const codes = const <int>[38, 38, 40, 40, 37, 39, 37, 39, 66, 65];
```
brianegan: Rando Q, what do you think about something like this (dunno if it's better, was just a thought):

```dart
const up = 38;
const down = 40;
const left = 37;
const right = 39;
const a = 65;
const b = 66;

const konamiKeyCodes = [up, up, down, down, left, right, left, right, b, a];
```

web/konami/konamicode.dart
```diff
@@ -1,39 +1,28 @@
...
+  final result = querySelector('#result');
+  final keyboardInput$ctrl = new StreamController<KeyboardEvent>();
```
brianegan: I would personally argue against `$` "Finnish notation", but I know that's totally subjective. Perhaps we could chat about using this type of notation on gitter? If these are the canonical examples, these patterns could be copied, and I think it'd be good to quickly just figure naming out as a group so we can remain consistent. Whatcha think?

web/konami/konamicode.dart
```diff
@@ -1,39 +1,28 @@
...
+  final keyboardInput$ = new Observable(keyboardInput$ctrl.stream);
 
-  document.addEventListener('keyup', (Event event) => controller.add(event));
+  // Listen to the DOM for the 'keyup' event
+  // and it to the streamController each time it occurs
+  document.addEventListener('keyup', (event) => keyboardInput$ctrl.add(event));
```
brianegan: Not sure if this might hide what's going on / be a bit dense to folks, but you could pass the `add` method directly as the second argument:

```
document.addEventListener('keyup', keyboardInput$ctrl.add);
```

web/konami/konamicode.dart
dotdotcommadot: Good point! Will do!

web/konami/konamicode.dart
dotdotcommadot: Yeah, for me it doesn't matter how we name them, but it's definitely a good idea to keep it consistent.
Let's move it over to the Gitter chat :)

web/konami/konamicode.dart
dotdotcommadot: Jup!

web/konami/konamicode.dart
dotdotcommadot: Already thought of this one as well :)

frankpepermans:
Looking good!

Also check this #57

brianegan:
Looks great :) Nice work!!!

PR #56, opened 2017-02-07
fibonacci example

frankpepermans:
Fibonacci, rx-ish

see: http://introcs.cs.princeton.edu/java/23recursion/Fibonacci.java.html
and: https://gist.github.com/LambdaSix/b2dabdfe781cf7344f14


frankpepermans:
Did a small update, this removes Types within the `main()` function.

examples/bin/fibonacci.dart
```diff
@@ -0,0 +1,51 @@
...
+  /// initial n[1,2] values
+  Iterable<IndexedPair> start =
+      n == 0 ? const [] : n == 1 ? const [seq1] : const [seq1, seq2];
+
+  /// given that the first 2 numbers in the sequence don't need calculation...
+  var m = n >= 2 ? n - 2 : 0;
+
+  Observable
+      .range(0, m)
+      .skip(1)
```
brianegan: Hrm, do we need to do `range(0, m)` then `skip(1)`? Could just do up a `range(1, m)`, no?

frankpepermans:
You're right, the skip messy,
removed and simplified some other things!

dotdotcommadot:
Small point for consistency: maybe better to use the '//' instead of '///' as Brian pointed out earlier.
'///' is for documenting members and types
-> https://www.dartlang.org/guides/language/effective-dart/documentation#do-format-comments-like-sentences

PR #58, opened 2017-02-08
Updates DragDrop example to use document.onKeyUp stream

dotdotcommadot:


web/dragdrop/dragdrop.dart
```diff
@@ -1,53 +1,32 @@
...
-  Observable<MouseEvent> mouseMove = _getMouseObservable('mousemove');
-  Observable<MouseEvent> mouseDown = _getMouseObservable('mousedown');
+  final dragTarget = querySelector('#dragTarget');
+  final mouseUp = new Observable(document.onMouseUp);
+  final mouseMove = new Observable(document.onMouseMove);
+  final mouseDown = new Observable(document.onMouseDown);
 
   mouseDown
-    // Use map() to Calculate the left and top properties on mouseDown
+    /// Use map() to calculate the left and top properties on mouseDown
```
brianegan: I think the 3 `///` are only used for DartDocs. Therefore, this should be fine with just `//`.

web/dragdrop/dragdrop.dart
```diff
@@ -1,53 +1,32 @@
...
     .map((MouseEvent e) => <String, int>{
```
brianegan: Rather than a Map<String, int>, I think it would be best to use `dart:html`'s built in `Point` class.

dotdotcommadot: Good 'Point' !

PR #59, opened 2017-02-08
Added a 'Search GitHub' example

dotdotcommadot:


web/search_github/search_github.dart
```diff
@@ -0,0 +1,48 @@
...
+  keyUp
+    // Use map() to take the value from the input field
+    .map((event) => (event.target as InputElement).value)
+    // Use distinct() to ignore all keystrokes that don't have an impact on the input field's value (brake, ctrl, shift, ..)
+    .distinct()
+    // Use debounce() to prevent calling the server on fast following keystrokes
+    .debounce(new Duration(milliseconds: 250))
+    // Use call(onData) to clear resultsField
+    .call(onData: (_) => resultsField.innerHtml = '')
+    // Use flatMapLatest() to call the gitHub API
```
brianegan: Not sure how to word this in a succinct way, hehe, but I think it would also be cool to point out that `flatMapLatest` will only emit the latest search result, so you'll always get the correct results without having to manage anything, such as aborting the ajax requests.

Esp. if you're coming from futures, this is very nice :)

PR #65, opened 2017-02-19
Flutter example

brianegan:
A Flutter version of the Search example. Had a bit of extra fun and made it a bit more dynamic, depending on whether there are errors or not.

I plan on making the Github API class a shared component between the web and Flutter demo, but need Flutter to adopt the newest http package (should happen this week) before that can happen.

demo


dotdotcommadot:
Awesome job!! Will have an in-depth look at it tomorrow! :)

PR #66, opened 2017-02-21
Convert all transformer functions to Classes

brianegan:
As it stands, we currently export roughly 25 or 30 custom transformers for use, stand-alone from the base Observable class.

In order to use these transformers, you need to import and call a factory function, such as `maxTransformer(...)`. These functions return a new instance of a StreamTransformer. However, as Dart is such an OO language, constructing the transformers this way has always felt like a code smell to me, and I wonder if it makes more sense for this lib to export proper classes as opposed to static functions? The update would be quite trivial, e.g.

Before: `maxTransformer(compare)`
After: `new MaxStreamTransformer(compare)`

The actual change to the transformers themselves is also quick -- simply wrap the current function in a class, and use it to create the StreamTransformer in the Class constructor.

Let me know what you think! Happy to do this as a quick refactor if you think it's valuable.

For now, I've converted the MaxStreamTransformer to give you a feel for what it would look like.


brianegan:
Also address the final comment of https://github.com/ReactiveX/rxdart/pull/70 -- removes `retry` operator.

PR #70, opened 2017-02-23
rework retry

frankpepermans:
Up for discussion!

In this version, it asserts to ensure that the provided Stream is a broadcast Stream.

If it isn't, a resubscribe is impossible, making the retry implementation invalid.

See also: http://reactivex.io/documentation/operators/retry.html

### DIFF
--- a/lib/src/observable.dart
+++ b/lib/src/observable.dart
@@ -1040,7 +1040,7 @@ class Observable<T> extends Stream<T> {
   /// Creates an Observable that will repeat the source sequence the specified
   /// number of times until it successfully terminates. If the retry count is
   /// not specified, it retries indefinitely.
-  Observable<T> retry([int count]) => transform(retryTransformer(count));
+  Observable<T> retry({int count: 0}) => transform(retryTransformer(count: count));
 
   /// Creates an Observable that will emit the latest value from the source
   /// sequence whenever the sampleStream itself emits a value.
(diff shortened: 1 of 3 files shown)


brianegan:
Hey, good to get this convo started! After thinking about it a bit, I'm worried we can might not be able to use this solution :(

Totally correct me if I'm wrong, but from what I understand, when you subscribe to a broadcast stream a second time, it doesn't quite act like an Rx Observable. Rather than starting the whole process over again (Rx style), you simply subscribe to a "live" stream that may emit *more* values, but it won't start the whole operation from scratch (e.g. trying to make a new API call every time, for example).

Therefore, what we really need is a single-subscription Stream Factory. That would allow us to create & recreate the Stream as many times as the user passes in as a param. I could see this taking a couple forms:

  1. Retry might need to be Stream rather than a StreamTransformer. It would take a StreamFactory (similar to the `defer` constructor), and use that.

  2. See if we can find a way to clone streams internally, so the user doesn't have to think about this.

What are your thoughts? Would love to hear @lestathc opinion on this as well if he has time!

frankpepermans:
+1 for option 1 :)

lestathc:
Agree with @brianegan 

I replied my thoughts to that bug.

frankpepermans:
Yup, we've also discussed the factory type solution over at Gitter (feel
free to join there, much easier to conversate!)

I propose to temporarily remove the retry operator until a good solution
can replace it.

brianegan:
+1. I've done that in my rework of the Stream Transformer classes.

PR #72, opened 2017-02-26
Create Retry Factory

### LINKED ISSUE #69: Retry operator did not re-trigger onListen/onSubscribe side effect
As described in [ReactiveX `retry` operator](http://reactivex.io/documentation/operators/retry.html), current observable should be resubscribed once an error occurs. 

But existing implementation does not do that.

brianegan:
Fixes #69 

This is another attempt to fix retry. It is a Stream / Observable factory method that takes in a `streamFactory` and `count`. It then builds a stream from the factory, listens, and rebuilds / relistens the stream if an error occurrs.

I've tried to pay careful attention to the implementation, but as always, I'm sure I missed something :)

@frankpepermans  & @lestathc -- please have a look and let me know what you think!


frankpepermans:
Can't think of a better solution than having a factory :)

Works for me!

Issue #68, opened 2017-02-21
Does rxdart Observable want to follow the standard ReaxtiveX onError contract?


lestathc:
Hi, 

I am very interesting in this lib, this is a big improvement for Dart.

But I have a concern:

1. as described in the [contract of observable](http://reactivex.io/documentation/contract.html), the onError will terminate the subscription. 

2. but as described in the [Stream.listen](https://api.dartlang.org/stable/1.22.0/dart-async/Stream/listen.html), the `cancelOnError` is false (by default), subscription will still exist.

How does this lib want to deal this case?

Thanks

brianegan:
This is a good Q, and something we should document. Good catch, and certainly a tension between two worlds.

What do you think about this? Would love to hear your opinion.

(For what it's worth, I'd prefer to stay more in line with Dart Streams rather than Observables, as I want to fit into that ecosystem as smoothly as possible).

Overall, would love to hear your opinion, and will definitely add docs around this! Thanks for writing in!

lestathc:
I don't know the solution neither. This is not an implementation problem.

I agree with you that stay with Dart Stream (given it is in the Dart ecosystem) and document it. (but... I really really dislike the 'use with care' in document, like another [bug](https://github.com/ReactiveX/rxdart/issues/67) I posted to this lib, SyncStreamController could solve that but SyncStreamController is documented with ['it must be used carefully'](https://api.dartlang.org/be/144789/dart-async/SynchronousStreamController-class.html))

And if `cancelOnError` is false, what is the differences with the [`retry` operator](http://reactivex.io/documentation/operators/retry.html), user may be confused with it. (And also, I think the [retry in this lib](https://github.com/ReactiveX/rxdart/blob/master/lib/src/transformers/retry.dart) is not what the retry documented in the ReactiveX, this lib's retry did not **re-trigger** the onListen/onSubscribe side effect.)

There are also some other concepts are different between Stream and ReactiveX. Like:
* `first`, `last`... operators: in Stream it is a Future, but in ReactiveX it is a Observable;
* the ReactiveX Observable could be multi-subscribed, but non broadcasting Stream is not;
* not like ReactiveX, Stream exposed back pressure to the StreamSubscription (the `pause`, `resume` method), it needs more consideration when implementing operators.

I feel that there is a big gap between Stream and standard Observable, maybe more documents are needed.

frankpepermans:
Retry indeed is off and needs a rework, would you mind creating a separate
issue for that one?

[...]

PR #73, opened 2017-02-26
Proposal: `asObservable` method on returned Futures

brianegan:
One thing that is admittedly awkward about the Observable vs Stream divide (as noted in #68) is that `first`, `last`, and several other methods currently return a plain `Future`, which is incompatible with Observable, and requires an smelly wrapper. `Futures` in Dart can be easily converted to `Streams`, much like `Single` from `RxJava2` can be easily converted back to a normal `Observable` using the method `toObservable`. But that's not the case in RxDart! For example:

```
new Stream.fromIterable([1, 2, 3]).last.asStream().listen(print); // prints 3
```

But: what if RxDart Observables had the same thing!?

```
new Observable.fromIterable([1, 2, 3]).last.asObservable().listen(print); // prints 3
```

Well that's what I've done. Any methods that returned a plain ol' `Future` now return a new, beefy `Future` called `ConvertibleFuture` that adds one extra method: `asObservable`.

This means any method that returns a `Future` from the `Observable` class can now be transparently converted back to an observable using `asObservable`.

Note: I'm really not sure if this is a good idea, I just thought it was a bit clever and solved the problem of converting back and forth between `Future` and `Observable`. However, it may be confusing, esp. to new folks who may accidentally get the impression any `Future` could be converted.

The alternative, of course, is duplicating all the `Future` returning methods, such as `first`, with a new name, e.g. `firstAsObservable`, that itself returns an `Observable` instead of a `Future`. Not sure which is better, just thought this was a fun idea and went for it.

@frankpepermans  and @lestathc Lemme know whatcha think and I'll go one direction or the other!

### DIFF
--- a/lib/src/as_observable_future.dart
+++ b/lib/src/as_observable_future.dart
@@ -1,0 +1,34 @@
+import 'dart:async';
+
+import 'package:rxdart/rxdart.dart';
+
+/// A future that can be converted directly to an Observable using
+/// the `asObservable` method.
+///
+/// This class simply wraps a normal Future, providing one additional method
+/// for smoother interoperability with the Observable class.
+class ConvertibleFuture<T> implements Future<T> {
+  final Future<T> wrapped;
+
+  ConvertibleFuture(this.wrapped);
+
+  Observable<T> asObservable() => new Observable<T>.fromFuture(wrapped);
+
+  @override
+  Stream<T> asStream() => wrapped.asStream();
+
+  @override
+  Future<T> catchError(Function onError, {bool test(Object error)}) =>
+      wrapped.catchError(onError, test: test);
+
+  @override
+  Future<S> then<S>(FutureOr<S> onValue(T value), {Function onError}) => wrapped
+      .then(onValue, onError: onError);
+
+  @override
+  Future<T> timeout(Duration timeLimit, {onTimeout()}) =>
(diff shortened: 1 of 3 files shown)


frankpepermans:
I wonder if we can use covariant in this scenario, it is a new feature from Dart 1.22
http://news.dartlang.org/2017/02/dart-122-faster-tools-assert-messages.html

brianegan:
Oh, interesting... Although, I think I'm understanding it :P I don't quite see how `covariant` will help us here? Feel like it would be more useful from the `Stream` class itself?

frankpepermans:
I'm not sure though, I go from Future to Observable in my day to day work quite often. These futures are unrelated to the Stream or Observable class.

How about a kind of utils solution? Eg a function Observable <T> observableFromFuture <T>(Future <T> value)?

brianegan:
Totally open to hearing more about that use case, but I think it's slightly different than what I'm going for here -- In this PR, I'm testing ideas of how to fluently convert from `Future` back to `Observable` within the context of writing a chain, so things like this are easy:

`new Observable.fromIterable([1, 2, 3]).first.asObservable().flatMap(apiCall).last.asObservable().filter(somePredicate)`

As a silly example.

brianegan:
Discussed on Gitter -- gonna shut this thread down and open up a PR with `xAsObservable()` methods.

PR #75, opened 2017-02-27
Retry factory

### LINKED ISSUE #69: Retry operator did not re-trigger onListen/onSubscribe side effect

brianegan:
Fixes #69 

This is another attempt to fix retry. It is a Stream / Observable factory method that takes in a `streamFactory` and `count`. It then builds a stream from the factory, listens, and rebuilds / relistens the stream if an error occurrs.

I've tried to pay careful attention to the implementation, but as always, I'm sure I missed something :)

@frankpepermans  & @lestathc -- please have a look and let me know what you think!

-- 

p.s. the last version of this pr got closed due to a falty rebase, and I had to reopen another PR.

### DIFF
--- a/lib/src/streams/defer.dart
+++ b/lib/src/streams/defer.dart
@@ -1,15 +1,17 @@
 import 'dart:async';
 
-class DeferStream<T> extends Stream<T> {
-  final StreamProvider<T> streamProvider;
+import 'package:rxdart/src/streams/utils.dart';
 
-  DeferStream(this.streamProvider);
+class DeferStream<T> extends Stream<T> {
+  final StreamFactory<T> streamFactory;
+
+  DeferStream(this.streamFactory);
 
   @override
   StreamSubscription<T> listen(void onData(T event),
       {Function onError, void onDone(), bool cancelOnError}) {
     StreamController<T> controller;
-    Stream<T> stream = streamProvider();
+    Stream<T> stream = streamFactory();
     bool hasListened = false;
 
     controller = new StreamController<T>(onListen: () {
@@ -23,5 +25,3 @@ class DeferStream<T> extends Stream<T> {
         onError: onError, onDone: onDone, cancelOnError: cancelOnError);
   }
 }
-
-typedef Stream<T> StreamProvider<T>();
--- a/lib/src/streams/defer.dart
+++ b/lib/src/streams/defer.dart
@@ -4,24 +4,17 @@ import 'package:rxdart/src/streams/utils.dart';
 
 class DeferStream<T> extends Stream<T> {
   final StreamFactory<T> streamFactory;
+  bool _isUsed = false;
 
   DeferStream(this.streamFactory);
 
   @override
   StreamSubscription<T> listen(void onData(T event),
       {Function onError, void onDone(), bool cancelOnError}) {
-    StreamController<T> controller;
-    Stream<T> stream = streamFactory();
-    bool hasListened = false;
+    if (_isUsed) throw new StateError("Stream has already been listened to.");
+    _isUsed = true;
 
-    controller = new StreamController<T>(onListen: () {
-      if (!hasListened) {
-        stream.listen(controller.add,
-            onError: controller.addError, onDone: controller.close);
-      }
-    });
-
-    return controller.stream.listen(onData,
+    return streamFactory().listen(onData,
         onError: onError, onDone: onDone, cancelOnError: cancelOnError);
   }
 }
--- a/lib/src/streams/retry.dart
+++ b/lib/src/streams/retry.dart
@@ -1,0 +1,79 @@
+import 'dart:async';
+
+import 'package:rxdart/src/streams/utils.dart';
+
+/// Creates a Stream that will recreate and re-listen to the source
+/// Stream the specified number of times until the Stream terminates
+/// successfully.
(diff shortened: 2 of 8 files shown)


lib/src/streams/retry.dart
```diff
@@ -0,0 +1,79 @@
...
+  void retry() {
+    bool errorOccurred = false;
+
+    subscription = streamFactory().listen((T data) {
+      controller.add(data);
+    }, onError: (dynamic e, dynamic s) {
+      errorOccurred = true;
+
+      if (count == retryStep) {
+        controller.addError(new RetryError(count));
```
frankpepermans: after addError, shouldn't we then also do `controller.close()`?

lib/src/streams/retry.dart
```diff
@@ -0,0 +1,79 @@
...
+  StreamSubscription<T> subscription;
+
+  RetryStream(this.streamFactory, [this.count]);
+
+  @override
+  StreamSubscription<T> listen(void onData(T event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    controller = new StreamController<T>(
+        sync: true,
+        onListen: retry,
```
lestathc: If the streamFactory returns a sync stream, this may potentially break
**An onListen callback from the controller is not an asynchronous event, and adding events to the controller in the onListen callback is always wrong. The events will be delivered before the listener has even received the subscription yet.** See [SynchronousStreamController](https://docs.flutter.io/flutter/dart-async/SynchronousStreamController-class.html)

Though I don't know why it is documented like that, and I just tried this piece of code and it works:
```dart
  StreamController sc;
  onListen() {
    sc.add(1);
  }
  sc = new StreamController(onListen: onListen, sync: true);
  
  sc.stream.listen(print);
}
```

lib/src/streams/retry.dart
```diff
@@ -0,0 +1,79 @@
...
+
+      if (count == retryStep) {
+        controller.addError(new RetryError(count));
+      } else {
+        subscription.cancel();
+        retryStep++;
+        retry();
+      }
+    }, onDone: () {
+      if (!errorOccurred) {
```
lestathc: can close directly since if subscription is cancelled, this code is not reached.

lib/src/streams/retry.dart
```diff
@@ -0,0 +1,79 @@
...
+///     new RetryStream(() {
+///          new Stream.fromIterable([1])
+///             .concatWith([new ErrorStream(new Error())]);
+///        }, 1)
+///        .listen(print, onError: (e, s) => print(e)); // Prints 1, 1, RetryError
+class RetryStream<T> extends Stream<T> {
+  final StreamFactory<T> streamFactory;
+  int count;
+  int retryStep = 0;
+  bool shouldClose = false;
```
lestathc: Seems `shouldClose` is not used?

lib/src/streams/retry.dart
```diff
@@ -0,0 +1,79 @@
...
+  StreamController<T> controller;
+  StreamSubscription<T> subscription;
+
+  RetryStream(this.streamFactory, [this.count]);
+
+  @override
+  StreamSubscription<T> listen(void onData(T event),
+      {Function onError, void onDone(), bool cancelOnError}) {
+    controller = new StreamController<T>(
```
lestathc: Given you did not override the `isBroadcastStream`, the `RetryStream` is single subscription, but if I listen it twice, the `controller` will be recreated and nothing happens.

lib/src/streams/retry.dart
brianegan: Hey, good Q! 

Yeah, if the stream is in the middle of processing an event, and you try to add *another* event in the `listen(onData)` callback, you'll run into errors.

I tried to a couple of different tests with different `StreamController` setups to see if this / how this breaks, and I didn't run into too many problems. I think we're Ok for now.

lib/src/streams/retry.dart
brianegan: Good catch, I'll fix that up!

lib/src/streams/defer.dart
```diff
@@ -1,15 +1,17 @@
...
+
 class DeferStream<T> extends Stream<T> {
-  final StreamProvider<T> streamProvider;
+  final StreamFactory<T> streamFactory;
 
-  DeferStream(this.streamProvider);
+  DeferStream(this.streamFactory);
 
   @override
   StreamSubscription<T> listen(void onData(T event),
```
brianegan: Good call, can probs simplify there.

brianegan:
Thanks for the comments @lestathc and @frankpepermans! I've address all feedback, let me know if you spot anything else!

lib/src/streams/retry.dart
lestathc: I remember that it do throw an exception in an older version of dart sdk. But this time I can not reproduce it again.

So, it is ok to keep it as is.

lib/src/streams/retry.dart
lestathc: I think all code in `streams/` folder may this fix as well.

PR #76, opened 2017-03-01
Ensure all Streams are single-subscription

brianegan:
### Problem

Our streams were not properly enforcing the single-subscription contract.

Many of our streams were "safe" from this bug due to the fact that they wrapped streams that threw a single-subscription error, but our streams did not explicitly fail on re-subscribe.

### Solution

Instead of re-creating the controller in `onListen` every time, therefore creating a fresh stream on every listen, we simply need to build the controller only once up front. That way, when the controller's stream is listened to a second time, it will be the same `stream` as the first attempt, and throw the appropriate error.

Please check it out and let me know whatcha think @frankpepermans  and @lestathc

### DIFF
--- a/lib/src/streams/amb.dart
+++ b/lib/src/streams/amb.dart
@@ -1,17 +1,24 @@
 import 'dart:async';
 
 class AmbStream<T> extends Stream<T> {
-  final Iterable<Stream<T>> streams;
+  final StreamController<T> controller;
 
-  AmbStream(this.streams);
+  AmbStream(Iterable<Stream<T>> streams)
+      : controller = _buildController(streams);
 
   @override
   StreamSubscription<T> listen(void onData(T event),
       {Function onError, void onDone(), bool cancelOnError}) {
+    return controller.stream.listen(onData,
+        onError: onError, onDone: onDone, cancelOnError: cancelOnError);
+  }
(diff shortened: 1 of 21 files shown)


frankpepermans:
Quick review looks ok to me, will go in depth tomorrow

frankpepermans:
Hmmm I think there is still a problem here,

now the controller is constructed right away, this breaks the whole lazy-initialization, take the following example:

I build an Observable like this:

```
var myObs = new Observable.just(1)
    .debounce(const Duration(seconds: 1));
```

At this point, the debounce already created an internal StreamController.

But what if the Stream is never listened to? The controller would then never close, causing a memory leak.

brianegan:
Discussed on Gitter, documenting here. Investigated class, found: "Had a quick look at the controller source code, it simply sends the done event and sets a flag to prohibit new events, but no special stuff"

Created a gist: https://gist.github.com/brianegan/9f3f8ad5281cbd1385f3ee0ecde6299b

This gist showed that "After the main closure exits, heap allocation drops back to about 0
During execution as the array is building it eventually gets to about 4 gigs on my machine."

Therefore, it appears all references are properly de-referenced, even if the controller hasn't been closed.

Issue #74, opened 2017-02-27
Scan/Min/Max transformer may has side effect leaking 


lestathc:
Considering I have such code:

```dart
var transformer = new ScanStreamTransformer(myPredictFunc);
var scan1 = myStream.transform(transformer);
var scan2 = myAnotherStream.transform(transformer);
```
I may dirty the intermedia result of scan transformer.

Given the transforms are public, if I only want to reuse the transforms without using the Observable interface, and according to ["It is good practice to write transformers that can be used multiple times."](https://api.dartlang.org/stable/1.22.1/dart-async/StreamTransformer-class.html). I may want to use transformer like that.

How about either mark transformers private, or document them/add annotations to indicate that these transformers is not sharable.

brianegan:
Hrm, good catch. I'd personally like to keep the transformers Public, as that allows folks to use all of our Stream and StreamTransformers classes individually without needing to pull in all of Observable if they really only need one or two items.

In this case, great call. We should inspect our operators and make sure they can be reused. Secondly, we defo need more docs on these classes. We've just made them public without much explanation thus far, something we'll certainly fix :)

frankpepermans:
hmmm, we *could* fix this by storing values in a Map for example,

take the min/max transformer, instead of storing a single `value` in the transformer class, we could create a `Map<EventSink, T>` instead, which stores values in relation to the provided Stream sink.

On close, we could delete the related `EventSink` key from that `Map`.

brianegan:
Another option: Now that we've got a class with an internal factory method, every time `bind` is called, we could build a new transformer internally and bind that, rather than building the transformer immediately upon creation and re-binding that. 

Although I think we should generally review our transformers and see if we can avoid doing that, for some cases it might make sense to ensure we don't have state problems.

PR #77, opened 2017-03-01
added tests for Transformer reusability (same instance), fixed transf…

frankpepermans:
Added tests for operators using `Transformers`

fixed some transformers,
those that use `fromHandlers `and store values need to be factory created on `bind`

in other cases, when just returning a new `StreamTransformer`, all values needed to be moved within that `StreamTransformer `body, some were sitting above being part of the `Transformer `class scope

### DIFF
--- a/test/transformers/call_test.dart
+++ b/test/transformers/call_test.dart
@@ -51,6 +51,24 @@ void main() {
     }));
   });
 
+  test('rx.Observable.call.reusable', () async {
+    bool onDataCalled = false;
+    final CallStreamTransformer<int> transformer =
+        new CallStreamTransformer<int>(onData: (_) {
+      onDataCalled = true;
+    });
+    final Observable<int> observableA = new Observable<int>.just(1);
+    final Observable<int> observableB = new Observable<int>.just(1);
+
+    observableA.transform(transformer).listen((_) {}, onDone: expectAsync0(() {
+      expect(onDataCalled, isTrue);
+    }));
+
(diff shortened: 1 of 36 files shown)


frankpepermans:
This relates to https://github.com/ReactiveX/rxdart/issues/74
@lestathc @brianegan

test/transformers/call_test.dart
```diff
@@ -51,6 +51,24 @@ void main() {
...
+  test('rx.Observable.call.reusable', () async {
+    bool onDataCalled = false;
+    final CallStreamTransformer<int> transformer =
+        new CallStreamTransformer<int>(onData: (_) {
+      onDataCalled = true;
+    });
+    final Observable<int> observableA = new Observable<int>.just(1);
+    final Observable<int> observableB = new Observable<int>.just(1);
+
+    observableA.transform(transformer).listen((_) {}, onDone: expectAsync0(() {
```
brianegan: Don't need to change the whole PR, but for this one might be good to use the new Stream matchers...

```
expect(observableA.transform(transformer), emitsInOrder(<dynamic>[1, emitsDone]));
```

frankpepermans:
heh dunno how I've missed those :D

I'll have a look at the Stream matchers later (work...), or we do it in a new PR?

brianegan:
Nope, I bumped our test version to the second-latest (latest test version has a bug, being fixed). 

So all these new matchers come from the test package as of a couple weeks ago!

PR #84, opened 2017-03-20
Change Max & Min to return a Future of the Max / Min value of a Stream

### LINKED ISSUE #83: Convert min / max to return single value
In order to maintain consistency with other Rx implementations, such as RxJS, we should return a single value from the max / min operators.

  - http://reactivex.io/documentation/operators/count.html
  - https://github.com/ReactiveX/rxjs/blob/master/src/operator/max.ts

brianegan:
As discussed on Gitter, fixes #83 -- Convert max/min to return a Future with a single value. This will help keep us more closely aligned with the other Rx implementations.

Changes:

  - README now points to master badge for tests & info about max / min
  - Removed old transformers
  - Created new Futures that can be used independently of the main lib
  - Added docs


brianegan:
Nothing we can do about the coverage here... All lines are covered, but since there's a bit less code to cover, the % of lines covered to uncovered went down slightly.

PR #86, opened 2017-03-30
Allow defer & subject streams to be listened to multiple times

brianegan:
Discussed on Gitter -- a current limitation of the Subject streams, having used DeferStreams, was that they could not be listened to multiple times properly, the way a broadcast stream should support.

This fixes that!

Proposal: extend DeferStream with an extra `reusable` option that can be used when the consumer is certain they're working with factory function that produces reusable streams, such as our subjects with broadcast streams.

Then, use this new capability in both subjects.

### DIFF
--- a/example/flutter/github_search/lib/github_search_api.dart
+++ b/example/flutter/github_search/lib/github_search_api.dart
@@ -1,6 +1,6 @@
 import 'dart:convert';
 
-import 'package:flutter/http.dart';
+import 'package:http/http.dart';
 import 'package:rxdart/rxdart.dart';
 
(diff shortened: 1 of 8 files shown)


example/flutter/github_search/lib/github_search_api.dart
```diff
@@ -1,6 +1,6 @@
 import 'dart:convert';
 
-import 'package:flutter/http.dart';
+import 'package:http/http.dart';
```
brianegan: Simple side-fix to the flutter example

Issue #91, opened 2017-04-23
Implement distinctUntilChanged


adaojunior:
Any plans to implement `distinctUntilChanged` ? see example on [rxjs](http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-distinctUntilChanged)

frankpepermans:
Dart has distinct, which is actually distinctUntilChanged, whereas the rxjs distinct is different and does not exist: https://api.dartlang.org/stable/1.19.1/dart-async/Stream/distinct.html

frankpepermans:
see also: https://github.com/ReactiveX/rxdart/pull/92

PR #92, opened 2017-04-23
DistinctUnique

frankpepermans:
It came to my attention that Rx distinct and Dart distinct are two different things.

In Rx, distinct will emit unique events only, whereas in Dart, it only compares the last event to the previous one and test for equality.

The Dart distinct is actually equal to the Rx distinctUntilChanged,
The Rx distinct currently has no Dart implementation, so I've decided to call it distinctUnique.

*confusing I know*

so 
- rx.Distinct == dart.distinctUnique
- rx.DistinctUntilChanged = dart.distinct

see:
- http://reactivex.io/documentation/operators/distinct.html
- https://api.dartlang.org/stable/1.23.0/dart-async/Stream/distinct.html
- http://www.introtorx.com/Content/v1.0.10621.0/05_Filtering.html#Distinct
- http://rxmarbles.com/#distinctUntilChanged


frankpepermans:
see also: https://github.com/ReactiveX/rxdart/issues/91

PR #100, opened 2017-06-14
Proper error handling

frankpepermans:
Added try/catch blocks where required
Added parameter checks, for example combineLatest([s1, s2, null]) => should throw because a null is passed instead of a Stream
Added more tests for Errors
Fixed old Error tests where expectAsync was missing
Removed call


lib/src/streams/amb.dart
```diff
@@ -36,24 +37,35 @@ class AmbStream<T> extends Stream<T> {
...
+              controller.add(value);
+            } catch (e, s) {
+              controller.addError(e, s);
+            }
           }
 
-          for (int i = 0, len = streams.length; i < len; i++) {
-            Stream<T> stream = streams.elementAt(i);
+          if (streams == null) {
+            controller.addError(new ArgumentError('streams cannot be null'));
```
brianegan: The ArgumentError cases are a great addition, but I don't think ArgumentErrors should be sent to the listener.

When errors occur within a FlatMap `mapper` function, for example, I'd expect those errors to be sent to the listener because the function can only be evaluated at a later point in time and you can't know if it will throw any errors until that later point in time.

However, when you pass invalid arguments into an operator which constructs a new stream, I think the ArgumentError should throw immediately. This is easier to debug because you'd have a stack trace leading directly to the actual problem. Additionally, it feels awkward to allow an operator to return a stream that we know WILL error and send no further data.

My recommendation: Move all `ArgumentErrors` out of the "Send to Listener" path and into an immediate throw at the top of the function.

lib/src/streams/combine_latest.dart
```diff
@@ -32,42 +32,51 @@ class CombineLatestStream<T> extends Stream<T> {
...
         sync: true,
         onListen: () {
-          final List<dynamic> values = new List<dynamic>(streams.length);
-          final List<bool> triggered =
-              new List<bool>.generate(streams.length, (_) => false);
-          final List<bool> completedStatus =
-              new List<bool>.generate(streams.length, (_) => false);
-          bool allStreamsHaveEvents = false;
+          if (streams == null) {
+            controller.addError(new ArgumentError('streams cannot be null'));
```
brianegan: I'd `throw` these ArgumentErrors at the top of the _buildController

lib/src/streams/concat.dart
```diff
@@ -36,25 +36,38 @@ class ConcatStream<T> extends Stream<T> {
...
-              index++;
-
-              if (index == len)
-                controller.close();
-              else
-                moveNext();
-            });
-          }
+          if (streams == null) {
+            controller.addError(new ArgumentError('streams cannot be null'));
```
brianegan: `throw` at top

lib/src/streams/concat_eager.dart
```diff
@@ -32,26 +32,41 @@ class ConcatEagerStream<T> extends Stream<T> {
...
+        streams != null ? new List<Completer<dynamic>>(streams.length) : null;
     StreamController<T> controller;
 
     controller = new StreamController<T>(
         sync: true,
         onListen: () {
-          for (int i = 0, len = streams.length; i < len; i++) {
-            completeEvents[i] = new Completer<dynamic>();
+          if (streams == null) {
+            controller.addError(new ArgumentError('streams cannot be null'));
```
brianegan: `throw` at top

lib/src/transformers/debounce.dart
```diff
@@ -38,13 +38,17 @@ class DebounceStreamTransformer<T> implements StreamTransformer<T, T> {
                 (T value) {
                   streamHasEvent = true;
 
-                  if (_timer != null && _timer.isActive) _timer.cancel();
+                  try {
+                    if (_timer != null && _timer.isActive) _timer.cancel();
 
-                  _timer = new Timer(duration, () {
-                    controller.add(value);
+                    _timer = Zone.current.createTimer(duration, () {
```
brianegan: Are these `Zone.current.createTimer` changes necessary / safe? Looks like the `new Timer` factory already does this `Zone.current.createTimer` binding for ya and has some extra stuff to make sure the execution is bound correctly to the Zone.

lib/src/streams/retry.dart
```diff
@@ -49,14 +49,16 @@ class RetryStream<T> extends Stream<T> {
   }
 
   void retry() {
-    subscription = streamFactory().listen((T data) {
-      controller.add(data);
-    }, onError: (dynamic e, dynamic s) {
-      subscription.cancel();
+    subscription = streamFactory().listen(controller.add,
+        onError: (dynamic e, dynamic s) async {
+      Future<dynamic> cancelFuture = subscription.cancel();
```
brianegan: Do we need these future variables? Can't we just do `await subscription.cancel()` and later on `await controller.close()`?

lib/src/transformers/buffer_with_count.dart
```diff
@@ -31,24 +31,28 @@ class BufferWithCountStreamTransformer<T, S extends List<T>>
   static StreamTransformer<T, S> _buildTransformer<T, S extends List<T>>(
       int count,
       [int skip]) {
-    final int skipAmount = ((skip == null) ? count : skip);
-
-    if (skipAmount <= 0 || skipAmount > count) {
```
brianegan: I'd keep these throws up here.

lib/src/transformers/interval.dart
```diff
@@ -25,10 +25,17 @@ class IntervalStreamTransformer<T> implements StreamTransformer<T, T> {
...
           onListen: () {
-            subscription = input.listen(
-                (T value) => subscription.pause(
-                    new Future<T>.delayed(duration, () => value)
-                        .then(controller.add)),
+            subscription = input.listen((T value) {
+              try {
+                final Completer<T> completer = new Completer<T>();
+
+                Zone.current.createTimer(duration, () => completer.complete(value));
```
brianegan: `new Timer`

PR #101, opened 2017-06-28
Fix   streams throw in ng2 after 0.13 update

frankpepermans:
Not sure why this was failing, with async on.

It may have to do with the await on cancel:
- parent stream adds event
- flat map latest receives it
- waits for the cancel of the previous subscription
- parent now closes
- flat map's onDone triggers, closing its controller
- the cancel from above is now complete, the listen function resumes
- Bad state error is now thrown (cannot add event after close)

seems plausible?

### DIFF
--- a/lib/src/transformers/flat_map_latest.dart
+++ b/lib/src/transformers/flat_map_latest.dart
@@ -39,14 +39,9 @@ class FlatMapLatestStreamTransformer<T, S> implements StreamTransformer<T, S> {
           sync: true,
           onListen: () {
             subscription = input.listen(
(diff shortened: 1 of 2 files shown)


brianegan:
Yep, I bet it's something like that. My guess is that by making it an async function, it's then put into the event loop queue to be executed at a later time (https://webdev.dartlang.org/articles/performance/event-loop).

Yet before the time comes for the function to execute, the `done` event is emitted, and our `close` function is executed synchronously, shutting down the stream. Then our function gets popped of the event queue and evaluated, resulting in a problem.

PR #102, opened 2017-06-28
Remove all `async` methods introduced in 0.13.0

brianegan:
These async methods were shown to introduce some subtle bugs that are very hard to debug and correct. I'd say we just get rid of em unless they're shown to fix an issue

### DIFF
--- a/test/transformers/flat_map_latest_test.dart
+++ b/test/transformers/flat_map_latest_test.dart
@@ -76,36 +76,6 @@ void main() {
     await expect(true, true);
   });
 
-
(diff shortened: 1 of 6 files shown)


test/transformers/flat_map_latest_test.dart
```diff
@@ -76,36 +76,6 @@ void main() {
     await expect(true, true);
   });
 
-
-  test('rx.Observable.flatMapLatest.error.shouldThrow12d122', () async {
```
brianegan: Accidental commit from before. Doh!

Issue #95, opened 2017-05-16
Should we call this RxDart or something else?


brianegan:
Overall, we want to make it easy for peeps coming from other languages to dive right into Dart, with all of the Rx power they're used to in other languages. We want to fit into the Dart ecosystem, and so we've based RxDart's `Observable` on Dart's `Stream` class. However, it's important to fundamentally understand Streams before diving into Observables, and there are some fundamental differences between `Streams` in Dart and `Observables` from the ReactiveX Contract. When those differences arise, we've decided the best thing to do is favor fitting into the Dart ecosystem over following the ReactiveX spec. 

Despite the differences between Dart Streams and Rx Observables, they're conceptually very similar, and we hoped calling it RxDart would make it easy for folks to get going quickly.

What we've seen, however, is a decent amount of confusion, and so we wanted to open a ticket and leave it open for a bit:

Should this library actually be a ReactiveX / RxDart library, or should we rename it to reinforce the fact that we are extending `Stream`, and that Streams are the most important thing to learn and understand? Has the name been helpful or confusing to you? Thanks for your feedback!

brianegan:
Would love to hear from @alexvbush, @lestathc and @adaojunior if y'all have time!

adaojunior:
When I first tried to use this library I got confused because I was trying to compare other implementations with this one, when I had a problem, I googled and usually I would get an answer with an example on JS/Java but it would be different on rxdart.
Overall, I believe extending `Stream` is more important to me than being 100% compatible with Rx.

zoechi:
I also think the long-term success depends on being close to Dart, while being close to Rx is mostly for a lower entry hurdle.

Perhaps there could be some helper classes that one doesn't need to use, but can use to get more Rx-like experience. 
I don't know if such an approach could work well though - just a thought.

alexvbush:
thank you for including me in this conversation! 

Ok, I'll be the bad guy here 😃  

As evident in this thread https://github.com/ReactiveX/rxdart/issues/88. In the beginning, I got confused just like @adaojunior. I would get examples in RxJS/RxSwift and try to do the same thing with RxDart but it would work differently.

At the end of the day I think that RxDart should be as close to canonical implementations of Rx (like RxJS, RxJava, etc.) as possible disregarding or bending where needed the dart stuff.

[...]

PR #103, opened 2017-08-07
PublishSubject added

pauldemarco:
Tests included.

### DIFF
--- a/lib/src/subject/publish_subject.dart
+++ b/lib/src/subject/publish_subject.dart
@@ -1,0 +1,150 @@
+import 'dart:async';
+import 'package:rxdart/src/observable.dart';
+
+/// A special StreamController that, once an Observer has subscribed,
+/// emits all subsequently observed items to the subscriber.
+///
+/// This subject allows sending data, error and done events to the listener.
+/// Any new events will be appropriately sent to the listeners.
+///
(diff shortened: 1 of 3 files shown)


brianegan:
First of all, thanks for the contribution! Overall, I'm totally open to this PR & the functionality it adds (it makes a lot of sense). I think it's good to simply chat about the history of this library a bit, and where we're coming from, so we can find a good solution / integration!

Since this library is based on Dart Streams, it can't comply 100% with the Rx Observable contract. We've had quite a few users write in, struggling with these differences. Therefore, we opened up an issue if we should rename this lib and make it very Dart-specific, and the consensus was that we SHOULD rename this library to something else: https://github.com/ReactiveX/rxdart/issues/95. 

Unfortunately, I think both @frankpepermans and I have been short on time lately and haven't had a chance to do the rename!

Due to this history, my personal feeling is that I'm a bit hesitant to add a new `Subject` / functionality that we plan on renaming / changing in the next month or so.

That said, I could totally see this fitting into our new lib, with a slightly different name but the same outcome. What are your thoughts, all?

brianegan:
Thanks for the work! I think we should just go ahead and add this for convenience to all and figure out the rename later.

That said, I thought it'd be good to do a quick refactor to reduce code duplication and couldn't push it atop your changes. Gonna close this out in favor of another PR.

PR #105, opened 2017-08-19
Publish subject

brianegan:
follow up to https://github.com/ReactiveX/rxdart/pull/103 with a refactor of the subject class.


brianegan:
Overall code was reduced, which is why we dropped a bit in coverage. No increase in missed code, just decreased overall number of lines covered.

PR #106, opened 2017-09-05
fixes an RTE when the returned cancel is Null instead of Future

frankpepermans:
Calling cancel() threw an RTE sometimes, since the cancel() method can return Null instead of Future

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,5 +1,5 @@
 name: rxdart
-version: 0.14.0
+version: 0.14.0+1
 authors:
 - Frank Pepermans <frank@igindo.com>
 - Brian Egan <brian@brianegan.com>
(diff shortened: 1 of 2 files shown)


pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: rxdart
-version: 0.14.0
+version: 0.14.0+1
```
brianegan: Should we go with 0.14.1 to maintain semver?

frankpepermans: I think 'plus' is allowed, it's only a simple hotfix in this case, see http://semver.org/#spec-item-10

brianegan: Yep, totally it's allowed, but I think the third "patch" digit is intended for fixes like this: "PATCH version when you make backwards-compatible bug fixes."

PR #107, opened 2017-09-06
Fix Test Failures

brianegan:
Getting flaky behavior on TravisCI, looks like timing failures. I've added more time between stream emissions to avoid these bugs.


brianegan:
That didn't work.

PR #108, opened 2017-10-09
Strong type improvement(s)

frankpepermans:
Just a small one, asyncMap was always returning Stream<dynamic> before, now it can infer the return type from the FutureOr Type (Stream<S> from FutureOr<S>)


brianegan:
Nice, looks good!

Issue #110, opened 2017-11-14
Looking for a way to have a stream with debounce cancelled when the base stream is closed


zoechi:
```dart
    _tickerSubscription = new Observable(onStateChange)
        .doOnDone(() {
          print('done: ${new DateTime.now().second}');
          _tickerSubscription.cancel();
        })
        .debounce(const Duration(seconds: 5))
        .listen((_) {
      print('event: ${new DateTime.now().second}');
        });
```

prints 
```
done: 56
00:00 [32m+29[0m: All tests passed![0m
event: 1
```

This is only an issue when I run it with
```bash
#!/usr/bin/env bash
dart --pause-isolates-on-exit --enable-vm-service=8899 test/all.dart &
sleep 1s
pub global run coverage:collect_coverage --uri=http://localhost:8899 -o coverage/coverage.json --resume-isolates
pub global run coverage:format_coverage --packages=.packages -i coverage/coverage.json -l -o coverage/lcov.info
```
With `pub run test` when "All tests passed!" is printed, there is no dart process hanging around anymore.

I use this in a test with built_redux where I call `store.dispose()` in `tearDown()`

zoechi:
With `--wait-paused` code coverage waits for the debounce timeout to complete, but at least doesn't keep the process open after it's done
```bash
pub global run coverage:collect_coverage --uri=http://localhost:8899 -o coverage/coverage.json --resume-isolates --wait-paused
```

brianegan:
@zoechi I think you've found a bug in our implementation of debounce. Just compared it to RxJS, and found the following:

  - When the base Stream ends, the debounced stream should immediately emit the last item.
  - If the subscription to the debounced stream is cancelled in doOnDone, the debounced stream should not emit any more values.

Does this sound like the behavior you're expecting? If so, I can write up a couple test cases + fix for it.

zoechi:
@brianegan yes, that sounds exactly like what I'm looking for.

PR #111, opened 2017-11-20
Debounce Fixes

brianegan:
Ensure debounce cancels any timers and emits the last item immediately after the source stream terminates.

Aims to solve #110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,8 @@ script:
   - pub run test test/rxdart_test.dart
   - pub get --packages-dir
   - pub global activate coverage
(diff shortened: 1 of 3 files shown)


zoechi:
This change causes most of my tests to fail. I wasn't able yet to track down why exactly

brianegan:
@zoechi Interesting -- all of the test cases we have are passing, so if you find something the bug, I'd be happy to accept changes on top of this branch with a test for the problem.

zoechi:
@brianegan @frankpepermans it's working fine now. 
I tracked the culprit down and it was easy to fix in my code.

PR #113, opened 2017-12-04
Updates examples using deprecated .call() method

dotdotcommadot:
Updates examples using the deprecated .call() method, with the newer .doOnData() method


brianegan:
Oh dang, good call!

PR #116, opened 2018-01-15
Implement exhaustMap from rxjs

brianegan:
ExhaustMap is like FlatMap, but will not produce a Streams if the previously created Stream is still emitting items.

Useful when you're calling async code based on a stream of scroll or mousemove events.

### DIFF
--- a/lib/src/transformers/exhaust_map.dart
+++ b/lib/src/transformers/exhaust_map.dart
@@ -1,0 +1,82 @@
+import 'dart:async';
+
+/// Converts items from the source stream into a new Stream using a given
+/// mapper. It ignores all items from the source stream until the new stream
+/// completes.
+///
+/// Useful when you have a noisy source Stream and only want to respond once
+/// the previous async operation is finished.
+///
+/// ### Example
+///     // Emits 0, 1, 2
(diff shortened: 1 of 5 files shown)


lib/src/transformers/exhaust_map.dart
```diff
@@ -0,0 +1,82 @@
...
+              try {
+                if (!outputIsEmitting) {
+                  outputIsEmitting = true;
+                  outputSubscription = mapper(value).listen(
+                    controller.add,
+                    onError: controller.addError,
+                    onDone: () {
+                      outputIsEmitting = false;
+                      if (inputClosed) controller.close();
+                    },
```
frankpepermans: Wondering if we need to add `cancelOnError: cancelOnError` here as well?
(note, we don't do this elsewhere though, see flatMapLatest for example).

It's true that the outer Stream would cancel as soon as the inner stream throws, if cancelOnError were true, but this is an asyn operation, if the inner stream does not cancel right away after the first Error, and it happens to be a sync Stream which receives a huge amount of events, then this would cause an unexpected performance hit (i.e. keep being bombarded by events as the code awaits the outer Stream to properly cancel)

Thoughts?

brianegan:
Included in #123.

PR #119, opened 2018-02-13
DDC 1.x prototype for discussion

robbecker-wf:
# Overview
My ham-fisted attempt to fix problems from compiling with DDC. We can use this PR for discussion or if it turns out to be good then, maybe we can eventually merge it.

The approach is to turn off implicit-casts and fix any type issues that the analyzer reports. To test that everything works, we can run the tests and the example/web with DDC and they should both work.

# Testing
 - [ ] `pub build --web-compiler=dartdevc example/web` succeeds
 - [ ] `pub serve --web-compiler=dartdevc example/web` the demos should all work
 - [ ] `dartanalyzer lib` passes with no errors

/cc @frankpepermans

### DIFF
--- a/.analysis_options
+++ b/.analysis_options
@@ -1,5 +1,7 @@
 analyzer:
-  strong-mode: true
+  strong-mode:
+    implicit-casts: false
+    implicit-dynamic: true
   exclude:
     - 'example/**'
 linter:
--- a/lib/src/futures/wrapped_future.dart
+++ b/lib/src/futures/wrapped_future.dart
@@ -25,7 +25,7 @@ class WrappedFuture<T> implements Future<T> {
       wrapped.then(onValue, onError: onError);
 
   @override
-  Future<T> timeout(Duration timeLimit, {onTimeout()}) =>
+  Future<T> timeout(Duration timeLimit, {FutureOr<T> onTimeout()}) =>
       wrapped.timeout(timeLimit, onTimeout: onTimeout);
 
   @override
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -90,7 +90,7 @@ abstract class Subject<T> implements StreamController<T> {
 
     source.listen((T event) {
       _add(event);
-    }, onError: (dynamic e, dynamic s) {
+    }, onError: (dynamic e, StackTrace s) {
       controller.addError(e, s);
 
       if (cancelOnError) {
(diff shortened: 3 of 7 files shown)


.analysis_options
```diff
@@ -1,5 +1,7 @@
 analyzer:
-  strong-mode: true
+  strong-mode:
+    implicit-casts: false
```
robbecker-wf: setting this to false makes the analyzer report errors where an implicit cast occurs.
Fixes usually are to be explicit with types (usually generic types) for Futures, Completers, Streams.

lib/src/subjects/subject.dart
```diff
@@ -90,7 +90,7 @@ abstract class Subject<T> implements StreamController<T> {
 
     source.listen((T event) {
       _add(event);
-    }, onError: (dynamic e, dynamic s) {
+    }, onError: (dynamic e, StackTrace s) {
```
robbecker-wf: According to the source in onError, the 2nd argument can be typed as StackTrace and it will either be there or be null.

lib/src/futures/wrapped_future.dart
```diff
@@ -25,7 +25,7 @@ class WrappedFuture<T> implements Future<T> {
       wrapped.then(onValue, onError: onError);
 
   @override
-  Future<T> timeout(Duration timeLimit, {onTimeout()}) =>
+  Future<T> timeout(Duration timeLimit, {FutureOr<T> onTimeout()}) =>
```
robbecker-wf: fairly certain this is ok. Just being explicit with the return type of onTimeout

lib/src/transformers/do.dart
```diff
@@ -154,8 +154,8 @@ class DoStreamTransformer<T> implements StreamTransformer<T, T> {
             );
           },
         );
-      };
-      final Function onCancelLocal = () {
+      }
+      void onCancelLocal() {
```
robbecker-wf: Changed to a full nested function instead of a local variable that is a Function

lib/src/transformers/do.dart
```diff
@@ -170,8 +170,8 @@ class DoStreamTransformer<T> implements StreamTransformer<T, T> {
         Future<dynamic> cancelFuture =
             subscriptions[input].cancel() ?? new Future<dynamic>.value();
 
-        return cancelFuture.whenComplete(() => subscriptions.remove(input));
-      };
+        cancelFuture.whenComplete(() => subscriptions.remove(input));
```
robbecker-wf: the function signature has void for a return type, so just remove the return. 🤷‍♂️ Should be ok?

brianegan:
Thanks so much for kicking this off and reporting! 

Gonna close it down in favor of #123 which includes these changes and makes the rest of the necessary changes :)

PR #123, opened 2018-02-16
final dart1 release, DDC with 1.24.3

frankpepermans:
Adds exhaustMap
Fixes code to work with the dartdevc
renamed .analysis_options to analysis_options .yaml
prepares release (readme, changelog, pubspec)


test/streams/zip_test.dart
```diff
@@ -285,12 +285,12 @@ void main() {
...
 
     observableWithError.listen(null,
-        onError: expectAsync2((dynamic e, dynamic s) {
+        onError: expectAsync2((Exception e, StackTrace s) {
       expect(e, isException);
     }));
   });
 
-  test('rx.Observable.zip.error.shouldThrowB', () {
+  /*test('rx.Observable.zip.error.shouldThrowB', () {
```
brianegan: Should we delete this test or get it working?

lib/src/transformers/do.dart
```diff
@@ -87,7 +90,7 @@ class DoStreamTransformer<T> implements StreamTransformer<T, T> {
 
     return new StreamTransformer<T, T>((Stream<T> input, bool cancelOnError) {
       StreamController<T> controller;
-      final Function onListenLocal = () {
+      final VoidFunc onListenLocal = () {
```
brianegan: No need to change it, but we can consider taking advantage of the new function syntax in upcoming PRs. For example, instead of a typedef, just type this with a `void Function()`

brianegan:
Haha, all right -- spent a couple hours trying to get the examples working using DDC, but not sure it's worth it!

Only change I'd propose to this code: Should we just keep dart2js as the default compiler for folks serving up the web examples? Seems a bit silly to serve up broken examples with DDC when they work perfectly fine with dart2js.

other than that, I'd say we should publish this version and get the Dart 2.0 changes lined up :)

robbecker-wf:
I suppose if the examples don't work, it makes sense to remove the webcompiler=dartdevc from the pubspec.

PR #124, opened 2018-02-21
fixes all current 2.0 issues

frankpepermans:
Initial changes, no analyzer warnings, works with ddc and dart2


pubspec.yaml
```diff
@@ -1,16 +1,15 @@
...
 description: RxDart is an implementation of the popular reactiveX api for asynchronous programming, leveraging the native Dart Streams api.
 homepage: https://github.com/ReactiveX/rxdart
 environment:
-  sdk: '>=1.21.0 <2.0.0'
-dependencies:
+  sdk: '>=2.0.0-dev <3.0.0'
 dev_dependencies:
   browser: ^0.10.0+2
-  test: 0.12.19
+  test: ^0.12.19
```
brianegan: This needs to be upgraded to at least `test: ^0.12.20` when `expectLater` was introduced. Maybe just go ahead and bump to the latest:   `test: ^0.12.30+3`

example/browser/pubspec.yaml
```diff
@@ -0,0 +1,16 @@
...
+authors:
+- Frank Pepermans <frank@igindo.com>
+- Brian Egan <brian@brianegan.com>
+description: Some examples which use Rx
+homepage: https://github.com/ReactiveX/rxdart
+environment:
+  sdk: '>=1.21.0 <2.0.0'
+dependencies:
+  rxdart:
+    path: ../../../rxdart
```
brianegan: If we're going to set the environment to `>=1.21.0 <2.0.0` (might actually need to make that >=1.24.0), should we point to a specific version of RxDart that works with 1.0, such as 0.15.1?

brianegan:
Rather than relying on `dev` and `stable` definitions in our travis-ci, we might need to specify specific versions for the time being to get these working

PR #135, opened 2018-03-28
added delay operator

frankpepermans:
Adds delay, see rxmarbles/rxjava/rxjs

### DIFF
--- a/lib/src/observable.dart
+++ b/lib/src/observable.dart
@@ -1133,6 +1133,22 @@ class Observable<T> extends Stream<T> {
   Observable<T> defaultIfEmpty(T defaultValue) =>
       transform(new DefaultIfEmptyStreamTransformer<T>(defaultValue));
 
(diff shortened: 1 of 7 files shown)


lib/src/observable.dart
```diff
@@ -1133,6 +1133,22 @@ class Observable<T> extends Stream<T> {
...
+  /// a particular increment of time (that you specify) before emitting
+  /// each of the source Observable’s items.
+  /// This has the effect of shifting the entire sequence of items emitted
+  /// by the Observable forward in time by that specified increment.
+  ///
+  /// [Interactive marble diagram](http://rxmarbles.com/#delay)
+  ///
+  /// ### Example
+  ///
+  ///     new Stream.fromIterable([1, 2, 3, 4])
```
brianegan: Example uses Stream + transformer, probs best to use an Observable + delay method here

PR #136, opened 2018-03-28
added bufferWithTimeframe

frankpepermans:
Added bufferWithTimeframe.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,10 +108,12 @@ var myObservable = Observable.combineLatest3(
     
 ##### Available Methods
 - [bufferWithCount](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferWithCount.html) / [BufferWithCountStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferWithCountStreamTransformer-class.html)
+- [bufferWithTimeframe](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferWithTimeframe.html) / [BufferWithTimeframeStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferWithTimeframeStreamTransformer-class.html)
 - [cast](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/cast.html) / [CastStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/CastStreamTransformer-class.html)
 - [concatMap](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/concatMap.html) (alias for `asyncExpand`)
 - [concatWith](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/concatWith.html)
 - [debounce](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/debounce.html) / [DebounceStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/DebounceStreamTransformer-class.html)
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -108,10 +108,12 @@ var myObservable = Observable.combineLatest3(
     
 ##### Available Methods
 - [bufferWithCount](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferWithCount.html) / [BufferWithCountStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferWithCountStreamTransformer-class.html)
+- [bufferWithTimeframe](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferWithTimeframe.html) / [BufferWithTimeframeStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferWithTimeframeStreamTransformer-class.html)
 - [cast](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/cast.html) / [CastStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/CastStreamTransformer-class.html)
```
brianegan: Just spotted this: Could we please remove cast now that we don't implement it ourselves?

lib/src/transformers/buffer_with_timeframe.dart
```diff
@@ -0,0 +1,84 @@
...
+
+            List<T> addToBuffer(T event) => buffer..add(event);
+
+            subscription = new Observable(input)
+                .doOnDone(() {
+                  doneController.add(true);
+                  doneController.close();
+                })
+                .map(addToBuffer)
+                .sample(new Observable<Null>.periodic(duration)
```
brianegan: I kinda think we should stick with Raw Streams + stream_transformers here, given the way we export our packages.

Right now, you can import only the transformers. If we use `Observable` here, and a user imports the `transformers` lib, it would now import all of the streams and observable stuff as well.

lib/src/transformers/buffer_with_timeframe.dart
```diff
@@ -0,0 +1,84 @@
...
+                    .takeUntil(doneController.stream))
+                .listen(
+                    (buffer) {
+                      controller.add(new List<T>.unmodifiable(buffer));
+                      buffer.clear();
+                    },
+                    onError: controller.addError,
+                    onDone: () {
+                      if (buffer.isNotEmpty) {
+                        scheduleMicrotask(() {
```
brianegan: Random Q: Why do we need to schedule this work on a microtask?

lib/src/transformers/buffer_with_timeframe.dart
```diff
@@ -0,0 +1,84 @@
...
+          onResume: () => subscription.resume(),
+          onCancel: () => subscription.cancel());
+
+      return controller.stream.listen(null);
+    });
+  }
+
+  static void assertTimeframe(Duration timeframe) {
+    if (timeframe == null) {
+      throw new ArgumentError('count cannot be null');
```
brianegan: Message should be: "timeframe cannot be null"

test/rxdart_test.dart
```diff
@@ -122,11 +127,13 @@ void main() {
   async_expand_test.main();
   async_map_test.main();
   buffer_with_count_test.main();
+  buffer_with_timeframe_test.main();
   concat_map_test.main();
   concat_with_test.main();
   contains_test.main();
   debounce_test.main();
   default_if_empty_test.main();
+  delay_test.main();
```
brianegan: Ah, was wondering why our coverage had dropped in the last PR. Good catch!

Issue #137, opened 2018-03-30
Rework buffer and window


frankpepermans:
Currently we have `bufferCount`/`windowCount` and `bufferTime`/`windowTime`, but we have no way to buffer on arbitrary intervals (for example, buffer on another Stream, or on a Future, ...)

Proposal:
We create a generic `buffer`/`window` operator which takes a `StreamSampler` implementation. 
This sampler is a simple interface which handles how the buffer builds and when the buffer should be emitted.

We provide a set of samplers that are commonly used:
- for `count`, [`skip`]: `OnCount`
- for `Duration`: `OnTime`
- for sampling on another `Stream`: `OnStream`
- for a recurring `Future`: `OnFuture`
- for a condition: `OnTest`

The user is free to create his/her own implementation `OnX`.

This will provide the following API:
```
observable
    .buffer(onTime(const Duration(seconds: 1)))
   .listen(...);

observable
    .buffer(onFuture(() => window.animationFrame()))
   .listen(...);

// stream emits: 1, 5, 20, 80, 120, 150, 200
observable
    .buffer(onTest((int event) => event > 100))
   .listen(...); // prints [1, 5, 20, 80, 120], [150, 200]
```
Which is equivalent to the following shorthand:
```
observable
    .bufferTime(const Duration(seconds: 1))
   .listen(...);

observable
    .bufferFuture(() => window.animationFrame())
   .listen(...);

// stream emits: 1, 5, 20, 80, 120, 150, 200
observable
    .bufferTest((int event) => event > 100)
   .listen(...); // prints [1, 5, 20, 80, 120], [150, 200]
```

brianegan:
I dig it. Nice abstraction that will allow us to create many different implementations without having to create a ton of new transformers

frankpepermans:
Pr merged!

PR #139, opened 2018-04-01
Samplers/buffer window

frankpepermans:
Rework buffer and window with onX samplers and the ability to create a custom sampler.

See https://github.com/ReactiveX/rxdart/issues/137


lib/src/transformers/window.dart
```diff
@@ -0,0 +1,77 @@
...
+          onPause: ([Future<dynamic> resumeSignal]) =>
+              subscription.pause(resumeSignal),
+          onResume: () => subscription.resume(),
+          onCancel: () => subscription.cancel());
+
+      return controller.stream.listen(null);
+    });
+  }
+
+  static void assertScheduler<T>(BufferBlocBuilder<T, Stream<T>> scheduler) {
```
brianegan: Hrm, I get whatcha mean with Scheduler, but might not be the best name since it has such a specific meaning in Rx

lib/src/transformers/buffer_test.dart
```diff
@@ -0,0 +1,60 @@
...
+
+/// Creates an Observable where each item is a list containing the items
+/// from the source sequence, sampled on a time frame.
+///
+/// ### Example
+///
+///     new Observable.periodic(const Duration(milliseconds: 100), (int i) => i)
+///       .bufferTime(const Duration(milliseconds: 220))
+///       .listen(print); // prints [0, 1] [2, 3] [4, 5] ...
+class BufferTestStreamTransformer<T> extends StreamTransformerBase<T, List<T>> {
```
brianegan: Probs don't need a reminder, but just to document Gitter chat: We decided to remove these in favor of 1 StreamTransformer for Buffer and Window

lib/src/transformers/window.dart
```diff
@@ -0,0 +1,77 @@
...
+/// Creates an Observable where each item is a list containing the items
+/// from the source sequence, sampled on a time frame.
+///
+/// ### Example
+///
+///     new Observable.periodic(const Duration(milliseconds: 100), (int i) => i)
+///       .bufferTime(const Duration(milliseconds: 220))
+///       .listen(print); // prints [0, 1] [2, 3] [4, 5] ...
+class WindowStreamTransformer<T> extends StreamTransformerBase<T, Stream<T>> {
+  final BufferBlocBuilder<T, Stream<T>> scheduler;
```
brianegan: Another Gitter Documentation: Change Bloc and go with Sampler

PR #140, opened 2018-04-02
Samplers/buffer window 2

brianegan:
Dummy PR so I can more easily review #139


README.md
```diff
@@ -107,8 +107,12 @@ var myObservable = Observable.combineLatest3(
 ### Transformations
     
 ##### Available Methods
+- [buffer](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/buffer.html) / [BufferStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferStreamTransformer-class.html)
 - [bufferCount](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferCount.html) / [BufferCountStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferCountStreamTransformer-class.html)
+- [bufferFuture](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/bufferFuture.html) / [BufferFutureStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx_transformers/BufferFutureStreamTransformer-class.html)
```
brianegan: Might be good to do a quick review of these links after removing the : Buffer* and Window* Transformers

lib/src/observable.dart
```diff
@@ -1095,21 +1138,56 @@ class Observable<T> extends Stream<T> {
...
+  /// Creates an Observable where each item is a [List] containing the items
+  /// from the source sequence, sampled on [onStream].
+  ///
+  /// ### Example
+  ///
+  ///     new Observable.periodic(const Duration(milliseconds: 100), (int i) => i)
+  ///       .bufferWhen(new Stream.periodic(const Duration(milliseconds: 220), (int i) => i))
+  ///       .listen(print); // prints [0, 1] [2, 3] [4, 5] ...
+  Observable<List<T>> bufferWhen(Stream<dynamic> other) =>
+      transform(new BufferStreamTransformer<T>(onStream(other)));
```
brianegan: I dig it. Everything using one nice abstraction!

lib/src/samplers/utils.dart
```diff
@@ -0,0 +1,16 @@
+import 'dart:async';
+
+typedef void OnDataTransform<T, S>(T event, EventSink<S> sink, [int skip]);
+
+/// An abstraction layer used internally by some [StreamTransformer]s
+///
+/// The bloc must implement a constructor which primarily takes a [Stream]
```
brianegan: bloc => Sampler

lib/src/samplers/utils.dart
```diff
@@ -0,0 +1,16 @@
...
+/// from the [StreamTransformer], together with other values if required.
+///
+/// It then transforms the data, to finally return it to the [StreamTransformer]
+/// via the [state] getter.
+///
+/// This is a generic interface that other bloc strategies can implement.
```
brianegan: bloc => sampler

lib/src/samplers/utils.dart
```diff
@@ -0,0 +1,16 @@
...
+abstract class OnSampler<T> {
```
brianegan: Question: Do you think this class is necessary? This looks a lot like a `StreamView`... could each of the `onX` functions simply return a Stream? That would change each of these class from `implements OnSampler` to `extends StreamView` with a call to `super(_stream)`?

That would also make `buffer` and `window` more flexible as they would accept a `SamplerBuilder` that returns any Stream?

PR #143, opened 2018-04-13
added Observable.fromCallback and .fromAsync

escamoteur:


lib/src/observable.dart
```diff
@@ -458,6 +458,26 @@ class Observable<T> extends Stream<T> {
...
       new Observable<T>(new DeferStream<T>(streamFactory, reusable: reusable));
 
+  /// Creates an Observable from a [callback] function.
+  /// Each time a new subscription is made the callback function is executed 
+  /// and its result emitted immideately  
+  ///   
+  /// The returned Observable is therefore a cold Observable.
+  /// 
+  factory Observable.fromCallback(T callBack()) =>
+      new Observable<T>.defer(() => new Observable<T>.just(callBack()), reusable: true);
```
brianegan: I actually think `reusable` should be `false` or optional. Streams are meant to be single-subscription, and this should only be used with a very specific intent.

lib/src/observable.dart
```diff
@@ -458,6 +458,26 @@ class Observable<T> extends Stream<T> {
...
+
+
+  /// Creates an Observable from  an asynchronous function [asyncFun] function.
+  /// Each time a new subscription is made the asynchronous function is called. 
+  /// Its result will be emited as soon as the function returns. 
+  ///   
+  /// The returned Observable is therefore a cold Observable.
+  /// 
+  factory Observable.fromAsync(Future<T> asyncFun()) =>
```
brianegan: Could we please add tests for both of these? You can see `DeferStream` as an example: https://github.com/ReactiveX/rxdart/blob/master/test/streams/defer_test.dart

lib/src/observable.dart
escamoteur: if we don't set reusable = false then it would not be possible to call it several times. But Frank just suggested to go a different way and use Stream generators which makes more sense

brianegan:
@frankpepermans @escamoteur Where did we land on this one?

escamoteur:
Forget it better to use genereraters

Issue #141, opened 2018-04-03
Make Subjects implement Observable interfaces


escamoteur:
Hi,

in other Rx implementations you can directly treat any Subject also as Observable instead of accessing it through an Observable field.

Would be nice if that could be changed.

Best
Thomas

brianegan:
PR Merged, will deploy as a new version once the Buffer and Window fixes are approved

PR #144, opened 2018-04-13
implemented interface of Observable on Subject

escamoteur:
all calls are relayed to subjects observable field.

Solves https://github.com/ReactiveX/rxdart/issues/141

### DIFF
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -1,5 +1,11 @@
 import 'dart:async';
+import 'package:rxdart/futures.dart';
 import 'package:rxdart/src/observable.dart';
+import 'package:rxdart/src/samplers/buffer_strategy.dart';
+import 'package:rxdart/src/transformers/time_interval.dart';
+import 'package:rxdart/src/transformers/timestamp.dart';
+import 'package:rxdart/src/utils/notification.dart';
+import 'package:rxdart/src/utils/type_token.dart';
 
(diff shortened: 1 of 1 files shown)


brianegan:
Oh dang, sorry to make you do all this! I think there might be an easier way to accomplish this task:

```dart
// Extend Observable
abstract class Subject<T> extends Observable<T> implements StreamController<T> {
  final StreamController<T> controller;
  final Observable<T> observable;
  bool _isAddingStreamItems = false;

  Subject(StreamController<T> controller, this.observable)
      : this.controller = controller,
        // Call super with the Observable
        super(observable);

   // Rest of Subject code here
}
```

escamoteur:
yeah, that's right but semantically a Subject is not derived from Observable though should only share an. interface,

brianegan:
Yah, I understand that feeling, but does the purity of implementing rather than extending buy us that much compared to the cost?

For example, we need to add tests for all of this code. This code and those tests are almost exact duplicates of the normal Observable class + tests.

In addition, every time we add something to Observable we now have to repeat it for Subject, including tests. It just feels like quite a large maintenance burden, and I'm not sure the "purity" is that much of a win when the practical effect in the end is the same.

Whatcha think?

escamoteur:
yea, that`s a point, I agree, Shit have thought about that earlier, will change it

escamoteur:
this line

```
  @override
  Observable<T> get stream => this;
```
 was too much :-) somehow it lead to an stack overflow

PR #145, opened 2018-04-13
Observable interface for subject second attempt

escamoteur:
Would be great if you could merge soon :-)

### DIFF
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -11,15 +11,15 @@ import 'package:rxdart/src/observable.dart';
 /// Please see `PublishSubject` for the simplest example of how to
 /// extend from this class, or `BehaviorSubject` for a slightly more
 /// complex example.
-abstract class Subject<T> implements StreamController<T> {
+abstract class Subject<T> extends Observable<T> implements StreamController<T> {
   final StreamController<T> controller;
   final Observable<T> observable;
   bool _isAddingStreamItems = false;
 
-  Subject(this.controller, this.observable);
+  Subject(this.controller, this.observable):super(observable);
 
   @override
-  Observable<T> get stream => observable;
+  Observable<T> get stream => this;
 
   @override
   StreamSink<T> get sink => new _StreamSinkWrapper<T>(this);
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -30,6 +30,9 @@ abstract class Subject<T> extends Observable<T> implements StreamController<T> {
   }
 
   @override
+  Observable<T> get stream => super.stream as Observable<T>;
+
+  @override
   ControllerCallback get onPause =>
       throw new UnsupportedError("Subjects do not support pause callbacks");
 
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -16,10 +16,7 @@ abstract class Subject<T> extends Observable<T> implements StreamController<T> {
   final Observable<T> observable;
   bool _isAddingStreamItems = false;
 
-  Subject(this.controller, this.observable):super(observable);
-
-  @override
-  Observable<T> get stream => this;
+  Subject(StreamController<T> controller, this.observable): this.controller = controller,super(observable);
 
   @override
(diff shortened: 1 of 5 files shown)


escamoteur:
I had to change one test because it expected to find an Observable and not a Stream. But I think that should not be a Problem. I expect most people in future to directly subscribe to the Subject and not accessing the .stream property

escamoteur:
https://github.com/ReactiveX/rxdart/issues/141

lib/src/subjects/subject.dart
```diff
@@ -11,15 +11,12 @@ import 'package:rxdart/src/observable.dart';
 /// Please see `PublishSubject` for the simplest example of how to
 /// extend from this class, or `BehaviorSubject` for a slightly more
 /// complex example.
-abstract class Subject<T> implements StreamController<T> {
+abstract class Subject<T> extends Observable<T> implements StreamController<T> {
```
frankpepermans: I'd change this to:
```
Subject(this.controller, Observable<T> observable) : super(observable);

  @override
  Observable<T> get stream => super.stream as Observable<T>;
```

lib/src/subjects/subject.dart
```diff
@@ -11,15 +11,12 @@ import 'package:rxdart/src/observable.dart';
...
   final StreamController<T> controller;
   final Observable<T> observable;
   bool _isAddingStreamItems = false;
 
-  Subject(this.controller, this.observable);
-
-  @override
-  Observable<T> get stream => observable;
```
frankpepermans: Keep the getter, but use:
```
@override
  Observable<T> get stream => super.stream as Observable<T>;
```

lib/src/subjects/subject.dart
```diff
@@ -11,15 +11,12 @@ import 'package:rxdart/src/observable.dart';
 /// Please see `PublishSubject` for the simplest example of how to
 /// extend from this class, or `BehaviorSubject` for a slightly more
 /// complex example.
-abstract class Subject<T> implements StreamController<T> {
+abstract class Subject<T> extends Observable<T> implements StreamController<T> {
   final StreamController<T> controller;
   final Observable<T> observable;
```
frankpepermans: remove this line, as Observable already exposes ```stream```

test/transformers/do_test.dart
```diff
@@ -30,7 +30,7 @@ void main() {
         () async {
       int count = 0;
       final BehaviorSubject<int> subject = new BehaviorSubject<int>(sync: true);
-      final Stream<int> stream = subject.stream.doOnError(
```
frankpepermans: You can leave the stream getter if you apply the above comments

escamoteur:
Good idea

lib/src/subjects/subject.dart
escamoteur: which line do you mean?

lib/src/subjects/subject.dart
escamoteur: Did this but now there is the warning we should avoid `as`

frankpepermans: Try making the stream property covariant in Observable

escamoteur: I fear I don't know how you mean that? I didn't want to touch the Observable class if possible

escamoteur: OK read now about what covariance means will try but if you could give me hint would be great

escamoteur: I now tried to find a solution for one hour without success. I give up.
The solution with `as` works. if you have a better idea please help me.

frankpepermans: Had a quick play with ```covariant```, but couldn't solve it cleanly either, we could just let ```Observable``` be as-is, and update ```Subject``` as following:
```
Subject(this.controller, Observable<T> observable)
      : this._observable = observable,
        super(observable);

  @override
  Observable<T> get stream => _observable;
```
Not sure if this is the cleanest solution though, I'll check again later on.

escamoteur: ok, will change it like that

brianegan:
LGTM

PR #152, opened 2018-05-22
OnErrorResumeNext updates

### LINKED ISSUE #142: onErrorResumeNext and onErrorReturn are missing error value
Would it be possible to use `onErrorResumeNext` and `onErrorReturn` with not only the value of item to emit, like this:
``` Dart
new Observable.error(new Exception())
  .onErrorReturn(1)
```
but also with a function that passes the error:
``` Dart
new Observable.error(new Exception())
  .onErrorReturn((error) => error is Exception ? 1 : 0 )
```
This functionality is super userful in other versions of Rx.

brianegan:
Fixes #142. 

  - Deprecate `OnErrorResumeNextStreamTransformer`
  - Add more flexible `OnErrorResumeStreamTransformer`
  - Add `onErrorResume` operator -- takes a function that returns a stream
  - Add `onErrorReturnWith` operator - takes a function that returns a final value
  - `onErrorResumeNext` and `onErrorReturn` operators use new transformer

### Questions

Since Dart can't overload a single method, I've used the following names:

  - `onErrorReturnWith`
  - `onErrorResume`
  - `OnErrorResumeStreamTransformer`

Please let me know if these sound good or if you have suggestions that might work better!

### DIFF
--- a/test/transformers/on_error_resume_test.dart
+++ b/test/transformers/on_error_resume_test.dart
@@ -60,18 +60,22 @@ void main() {
   });
 
   test('rx.Observable.onErrorResumeNext.pause.resume', () async {
+    final OnErrorResumeStreamTransformer<num> transformer =
+        new OnErrorResumeStreamTransformer<num>((dynamic e) => _getStream());
+    final List<num> exp = <num>[50] + expected;
     StreamSubscription<num> subscription;
     int count = 0;
 
-    subscription = new Observable<num>(new ErrorStream<num>(new Exception()))
-        .onErrorResumeNext(_getStream())
-        .listen(expectAsync1((num result) {
-          expect(result, expected[count++]);
+    subscription = new Observable<num>.merge(<Stream<num>>[
+      new Observable<num>.just(50),
+      new ErrorStream<num>(new Exception()),
+    ]).transform(transformer).listen(expectAsync1((num result) {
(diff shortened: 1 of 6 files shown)


test/transformers/on_error_resume_test.dart
```diff
@@ -0,0 +1,124 @@
...
+    int count = 0;
+
+    new Observable<num>(new ErrorStream<num>(new Exception()))
+        .onErrorResumeNext(_getStream())
+        .listen(expectAsync1((num result) {
+          expect(result, expected[count++]);
+        }, count: expected.length));
+  });
+
+  test('rx.Observable.onErrorResume', () async {
```
tomaszpolanski: Wouldn't it be beneficial to test also if the passed error is propagated properly? Something like this:
```
  test('rx.Observable.onErrorResume', () async {
    Exception exception = new Exception();
    new Observable<dynamic>(new ErrorStream<num>(exception))
        .onErrorResume((dynamic e) => new Observable<dynamic>.just(e))
        .listen(expectAsync1<dynamic, dynamic>((dynamic result) {
          expect(result, exception);
        }, count: 1));
  });
```

brianegan: Good call, I'll add that in. Thanks for the review!

PR #157, opened 2018-05-31
Fixed links for documentation in readme

charafau:
Many links were broken in README file so.. here's the fix. I hope I didn't miss single one.


brianegan:
Oh dang, DartDoc must have changed the way they're generating links. Thanks for fixing it up!

Issue #156, opened 2018-05-29
Inspect pause / resume behavior of Stream classes


brianegan:
Concat does not pause it's source streams in an `onPause` callback. Other Stream classes (concatEager, combineLatest, etc also fail to handle onPause / onResume).

We should look through these classes and properly implement pause / resume.

frankpepermans:
Argh :/ I'll take a look as well asap

frankpepermans:
track under https://github.com/ReactiveX/rxdart/pull/159

brianegan:
Fixed by #159

PR #159, opened 2018-06-06
implement missing onPause and onResume handlers on certain transformers

frankpepermans:
see https://github.com/ReactiveX/rxdart/issues/156

### DIFF
--- a/lib/src/streams/race.dart
+++ b/lib/src/streams/race.dart
@@ -66,6 +66,12 @@ class RaceStream<T> extends Stream<T> {
                 onError: controller.addError, onDone: () => controller.close());
           }
         },
+        onPause: ([Future<dynamic> resumeSignal]) => subscriptions
+            .where((subscription) => subscription != null)
+            .forEach((subscription) => subscription.pause(resumeSignal)),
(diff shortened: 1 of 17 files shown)


lib/src/streams/race.dart
```diff
@@ -66,6 +66,14 @@ class RaceStream<T> extends Stream<T> {
                 onError: controller.addError, onDone: () => controller.close());
           }
         },
+        onPause: ([Future<dynamic> resumeSignal]) => subscriptions
+            .where((StreamSubscription<T> subscription) => subscription != null)
```
brianegan: Is the null check necessary here?

frankpepermans:
I've added the ```pause```/```resume``` tests, plus I cleaned up ```amb```/```race``` a bit.
We had a fixed-length ```Subscription``` ```List``` before, once the race was over, all obsolete ```Subscription```s in that ```List``` were cancelled and then their reference in that List was set to ```Null```.

...Which is why there was a ```Null``` check in place when iterating over that ```List```...

Now I simply remove the obsolete ```Subscription```s from the ```List```, so only the active ```Stream``` remains in there.

PR #161, opened 2018-06-18
Add flatMapIterable, remove retype

### LINKED ISSUE #160: [error] The method 'retype' isn't defined for the class 'Stream<T>'
The deprecated `retype` method was removed as of Dart 2.0.0-dev.62.0 released June 13. Running tests in Chrome using `pub run build_runner test -- -p chrome` now fail at runtime with

    Error compiling dartdevc module:rxdart|lib/futures.ddc.js
  
    [error] The method 'retype' isn't defined for the class 'Stream<T>'. (package:rxdart/src/observable.dart, line 1909, col 33)

If you're not ready to remove the `retype` method from `rxdart` then maybe your implementation of `retype` can be explicitly call `cast`.

brianegan:
- Removes retype to fix #160.
  - Adds `flatMapIterable` operator (a simple but useful combo of FlatMap + Expand)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.18.0
+
+* Breaking Change: remove `retype` method, deprecated as part of Dart 2.
+* Add `flatMapIterable`
+
 ## 0.17.0
 
 * Breaking Change: `stream` property on Observable is now private.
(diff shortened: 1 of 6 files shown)


frankpepermans:
Looks good, will we release to pub? 
If so, we do still need a changelog/readme/version update

brianegan:
Good call! I'll add those changes and merge as a minor release

PR #176, opened 2018-08-10
Calling onDone after emit single value

SeniorIvan:
right behavior from http://reactivex.io/documentation/operators/just.html

### DIFF
--- a/lib/src/observable.dart
+++ b/lib/src/observable.dart
@@ -514,13 +514,22 @@ class Observable<T> extends Stream<T> {
 
   /// Creates an Observable that contains a single value
   ///
-  /// The value is emitted when the stream receives a listener.
+  /// The value is emitted when the stream receives a listener,
+  /// and done is called upon listening.
   ///
   /// ### Example
   ///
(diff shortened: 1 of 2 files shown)


frankpepermans:
Hey there,
Thanks for making a PR!

Is there an issue with the current implementation? You are forcing an async event here, which is not the right place, in Dart, Streams can be sync or async, depending in how you configure your StreamController.

An async Controller already uses scheduleMicrotask to add its events

brianegan:
Hey there -- since we use the `Stream.fromIterable` constructor under the hood, this isn't highly tested, but `Observable.just` currently emits `onDone` after the value is emitted.

For example, this test passes with the current implementation, and therefore I believe `.just` is exhibiting the behavior you're looking for:


```dart
test('rx.Observable.just', () {
  expect(new Observable<int>.just(1), emitsInOrder(<dynamic>[1, emitsDone]));
});
```

Like Frank, I'd be interested to hear if this isn't working for ya!

SeniorIvan:
see [https://github.com/ReactiveX/rxdart/issues/177](url)

PR #179, opened 2018-08-14
fix doOnError documentation.

dustin-graham:
the function that is called is provided an error and stacktrace objects.
If these are omitted by the developer as they follow the documentation,
we will end up with a NoSuchMethod unhandled exception and their doOnError
will never get called. This documentation should help the developer
understand how to implement this functionality correctly.


brianegan:
Oh, good call. Thanks!

PR #188, opened 2018-09-04
adds missing factory methods to readme

BugsBunnyBR:
Adds:
-  `empty`;
-  `eventTransformed`;
- `fromFuture`;
- `fromIterable`;

to the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ version 0.16.x is no longer backwards compatible and requires the Dart SDK 2.0
 
 ## How To Use RxDart
 
-### For Example: Reading the Konami Code 
+### For Example: Reading the Konami Code
 
 ```dart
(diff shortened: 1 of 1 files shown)


brianegan:
Thanks @BugsBunnyBR! Sorry for the delay -- was on vacation for a bit.

I wonder if it makes sense to put these in our documentation, since RxDart does not provide these constructors directly, but rather they come from the base Stream class? For example, we don't list all available Stream methods in our docs, only the StreamTranformers / operators that RxDart provides in addition to the Stream.

This makes me wonder whether we should really be listing the API methods in the README at all, or simply deferring to the API docs build by DartDoc for a complete reference.

@frankpepermans What are your thoughts on this PR?

Issue #183, opened 2018-08-24
Add SerialSubscription


sir-boformer:
`CompositeSubscription`
------

Basically a bag for multiple subscription (with `add` and `remove` methods) that can be unsubscribed at once: http://reactivex.io/RxJava/javadoc/io/reactivex/disposables/CompositeDisposable.html

```dart
final s = new CompositeSubscription();
s.add(firstStream.listen(...));
s.add(secondStream.listen(...));
//...
s.cancel(); // cancels all
```
It's very useful when you listen to multiple streams and want to dispose them all at once (e.g. when a widget in Flutter is disposed).

Danger of this one is that developers add thousands of subscriptions (e.g. for a repeatable task). In RxJava at least the "bag" is never emptied, even if the streams already emptied. 

`SerialSubscription`
---------

 A container that keeps track of a single subscription, and you can replace the subscription with another one (`set` method), which cancels the previous subscription: http://reactivex.io/RxJava/javadoc/io/reactivex/disposables/SerialDisposable.html

```dart
final s = new SerialSubscription();
s.set(firstStream.listen(...));
// ...
s.set(secondStream.listen(...)); // cancel firstStream subscription
// ...
s.cancel(); // cancel tracked subscription (if any)
```

Basically a shorthand for:

```dart
Subscription s;

s = firstStream.listen(...);
// ...
s?.cancel();
s = secondStream.listen(...);
// ...
s?.cancel();
```

I love this one. It ensures that subscriptions never get lost when you replace them. And you can also add them to `SerialSubscription`s.

Add `subscription` argument to `doOnListen` operator
----------

Use case: Easily add/set a subscription without breaking the flow:

```dart
final s = new CompositeSubscription();

observable
  .doOnListen(s.add)
  .listen(...);
```

sir-boformer:
What do you think? If it makes sense, I may take up the task.

brianegan:
Hey there :) Yah, we've had this request a couple of times. I think we've always just said "You can do it manually." E.g. CompositeSubscription could be done like so `List<StreamSubscription> subs = []; subs.forEach((sub) => sub.cancel());`). But maybe it's not the best answer since this question has been asked a few times now.

I'd be open to this contribution. If you take it up, could you please ensure you write tests for each? Thanks!

brianegan:
CompositeSubscription now added. Changing the title.

shtse8:
how about `SerialSubscription`?

PR #191, opened 2018-10-07
Add CompositeSubscription container to manage multiple subscriptions at once

Mishkun:
This PR is intended to tackle the issue #183.

It adds `CompositeSubscription`, which is (quoting the issue)
>Basically a bag for multiple subscription (with add and remove methods) that can be unsubscribed at once: http://reactivex.io/RxJava/javadoc/io/reactivex/disposables/CompositeDisposable.html
```dart
final s = new CompositeSubscription();
s.add(firstStream.listen(...));
s.add(secondStream.listen(...));
//...
s.cancel(); // cancels all
```
>It's very useful when you listen to multiple streams and want to dispose them all at once (e.g. when a widget in Flutter is disposed).

I've had a positive experience with composites in Android dev with RxJava, so I'm glad to bring this experience to dart users. But I have no experience with `Serial` variant and don't really know any usecases for that

### DIFF
--- a/lib/src/utils/composite_subscription.dart
+++ b/lib/src/utils/composite_subscription.dart
@@ -1,0 +1,68 @@
+
+import 'dart:async';
+
+/**
+ * Acts as a container for multiple subscriptions that can be canceled at once 
+ * e.g. view subcriptions in Flutter that need to be canceled on view disposal
+ * 
+ * Can be cleared or disposed. When disposed, cannot be used again.
+ * ### Example
+ *      // init your subscriptions
+ *      composite.add(observable1.listen(listener1))
+ *            ..add(observable2.listen(listener1))
+ *            ..add(observable3.listen(listener1));
+ * 
+ *      // clear them all at once
+ *      composite.clear();
+ */
+class CompositeSubscription {
+
+  bool _isDisposed = false;
+
+  final List<StreamSubscription> _subscriptionsList = new List<StreamSubscription>();
+
+  /**
+   * Checks if this composite is disposed. If it is, the composite can't be used again
+   * and will throw an error if you try to add more subscriptions to it.
+   */
+  bool get isDisposed => _isDisposed;
+
+  /**
+   * Adds new subscription to this composite.
+   * 
+   * Throws an exception if this composite was disposed
+   */
+  CompositeSubscription add(StreamSubscription subscription) {
+    if (isDisposed) throw("This composite was disposed, try to use new instance instead");
+    _subscriptionsList.add(subscription);
+    return this;
+  }
+
+  /**
+   * Removes subscription from this composite.
+   */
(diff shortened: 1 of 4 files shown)


lib/src/utils/composite_subscription.dart
```diff
@@ -0,0 +1,56 @@
...
+  CompositeSubscription add(StreamSubscription<dynamic> subscription) {
+    if (isDisposed)
+      throw ("This composite was disposed, try to use new instance instead");
+    _subscriptionsList.add(subscription);
+    return this;
+  }
+
+  /// Removes subscription from this composite.
+  CompositeSubscription remove(StreamSubscription<dynamic> subscription) {
+    _subscriptionsList.remove(subscription);
```
brianegan: Should this also cancel the subscription? Seems to do so in the RxJava implementation...

https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/disposables/CompositeDisposable.java#L153

Mishkun: Overlooked that thing, I'll change the code and test

lib/src/utils/composite_subscription.dart
```diff
@@ -0,0 +1,57 @@
...
+  bool get isDisposed => _isDisposed;
+
+  /// Adds new subscription to this composite.
+  /// 
+  /// Throws an exception if this composite was disposed
+  CompositeSubscription add(StreamSubscription<dynamic> subscription) {
+    if (isDisposed)
+      throw ("This composite was disposed, try to use new instance instead");
+    _subscriptionsList.add(subscription);
+    return this;
```
brianegan: Hey, was just thinking about this API a bit more and had one Q: Rather than returning `this`, should we return the `StreamSubscription` instead? This would allow users to write code like this:


```dart
final subscription = composite.add(myStream.listen(null));

composite.remove(subscription);
```

Rather than this:

```dart
final subscription = myStream.listen(null);
composite.add(subscription);

subscription.remove(subscription);
```

Chaining is still supported just fine via the cascade operator, e.g.

```dart
final subscription = CompositeSubscription()..add()..add()..add();
```

Mishkun: idk, In my experience of working with RxJava, you'd really rarely use the `remove` method.
When you are using composite, I think you already don't want to manage subscriptions in person.

frankpepermans: Returning *this* in Dart is indeed considered an anti-pattern:
https://www.dartlang.org/guides/language/effective-dart/design#avoid-returning-this-from-methods-just-to-enable-a-fluent-interface

Mishkun: Yeah, _"return this"_ thing is my my Java bad habit. I'll fix that

brianegan:
Looks great, thanks for contributing!

PR #195, opened 2018-10-13
Distinguish between seeded- and not-seeded BehaviorSubject

### LINKED ISSUE #170: Testing of BehaviorSubject with seed or first `null` value
I can't seem to test properly that `BehaviorSubject` emits `null` value first, either with `add` method or with seed value. I'm not sure if this is I'm doing wrong, but here's what I have:

```dart
test("Can expect initial non-null value", () {
  final subject = BehaviorSubject<String>(seedValue: "value");

  expect(subject, emits("value"));

  subject.close();
});

test("Can expect initial null value", () {
  final subject = BehaviorSubject<String>(seedValue: null);

  expect(subject, emits(null));

  subject.close();
});
```

For the non-null value the test passes. For the `null` value, the assertion fails because the stream is closed. If the stream is not closed, the test timeouts.

I feel something is off with `null` values being pushed as initial or first values to the subject, for example this:

```dart
test("Print initial values", () {
  BehaviorSubject<String>(seedValue: null)
    ..listen(print)
    ..add(null)
    ..add(null)
    ..close();
});
```

will print `null` only twice. When I replace nulls with instances, the value is printed three times. 

Overall seems like the existing `null` value isn't pushed to new subscribers. I haven't found it documented that for `null` it'd behave differently, and also I haven't found any information about special handling of `null` for RxDart in general.

lwasyl:
An attempt to fix #170 

As you can see, this prompted some more changes, specifically in `ValueConnectableObservable` and `Observable` itself. I wasn't quite sure how to handle them, so I'm looking forward to feedback.

Making this PR convinced me that `null` values should be straight away forbidden everywhere ;)

I haven't updated dartdoc or documentation yet, as I'm not sure the PR will be accepted. If the changes look good please let me know, and I'll push an update with documentation updated

### DIFF
--- a/lib/src/observables/connectable_observable.dart
+++ b/lib/src/observables/connectable_observable.dart
@@ -97,9 +97,20 @@ class ValueConnectableObservable<T> extends ConnectableObservable<T>
 
   ValueConnectableObservable._(this._source, this._subject) : super(_subject);
 
-  factory ValueConnectableObservable(Stream<T> source, {T seedValue}) =>
+  factory ValueConnectableObservable(Stream<T> source) =>
       ValueConnectableObservable<T>._(
-          source, BehaviorSubject<T>(seedValue: seedValue));
+        source,
+        BehaviorSubject<T>(),
+      );
+
+  factory ValueConnectableObservable.seeded(
+    Stream<T> source,
+    T seedValue,
+  ) =>
+      ValueConnectableObservable<T>._(
+        source,
+        BehaviorSubject<T>.seeded(seedValue),
+      );
 
   @override
   ValueObservable<T> autoConnect({
@@ -144,6 +155,9 @@ class ValueConnectableObservable<T> extends ConnectableObservable<T>
 
   @override
   T get value => _subject.value;
+
+  @override
+  bool get hasValue => _subject.hasValue;
 }
 
 /// A [ConnectableObservable] that converts a single-subscription Stream into
--- a/lib/src/observables/observable.dart
+++ b/lib/src/observables/observable.dart
@@ -2422,8 +2422,43 @@ class Observable<T> extends Stream<T> {
   /// // BehaviorSubject
   /// subscription.cancel();
   /// ```
-  ValueConnectableObservable<T> publishValue({T seedValue}) =>
-      ValueConnectableObservable<T>(this, seedValue: seedValue);
+  ValueConnectableObservable<T> publishValue() =>
+      ValueConnectableObservable<T>(this);
+
+  /// Convert the current Observable into a [ValueConnectableObservable]
+  /// that can be listened to multiple times, providing an initial seeded value.
+  /// It will not begin emitting items from the original Observable
+  /// until the `connect` method is invoked.
+  ///
+  /// This is useful for converting a single-subscription stream into a
+  /// broadcast Stream that replays the latest emitted value to any new
+  /// listener. It also provides access to the latest value synchronously.
+  ///
+  /// ### Example
+  ///
+  /// ```
+  /// final source = Observable.fromIterable([1, 2, 3]);
+  /// final connectable = source.publishValueSeeded(0);
+  ///
+  /// // Does not print anything at first
+  /// connectable.listen(print);
+  ///
+  /// // Start listening to the source Observable. Will cause the previous
+  /// // line to start printing 0, 1, 2, 3
+  /// final subscription = connectable.connect();
+  ///
+  /// // Late subscribers will receive the last emitted value
+  /// connectable.listen(print); // Prints 3
+  ///
+  /// // Can access the latest emitted value synchronously. Prints 3
+  /// print(connectable.value);
+  ///
+  /// // Stop emitting items from the source stream and close the underlying
+  /// // BehaviorSubject
+  /// subscription.cancel();
+  /// ```
+  ValueConnectableObservable<T> publishValueSeeded(T seedValue) =>
+      ValueConnectableObservable<T>.seeded(this, seedValue);
 
   /// Convert the current Observable into a [ReplayConnectableObservable]
   /// that can be listened to multiple times. It will not begin emitting items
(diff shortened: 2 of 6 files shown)


frankpepermans:
@brianegan hehe same here! I'll try to take a look asap tho

lwasyl:
No worries!

> want to make sure we get this change right so we don't need to make any more breaking API changes

If such breaking change is allowed, I'd disallow emitting `null`s, like RxJava 2, instead of merging this. While implementing this I realized how much complexity emitting nulls add. That said, it'd be a pretty big breaking change

brianegan:
Truly sorry about the delay here.. @lwasyl I'm actually totally with you. I actually think we should either not allow null at all (including rejecting them if folks try to `add` them).

I really appreciate the work you put in here! However, as I said on the original thread: I just don't think it's a great idea to change the constructor one of the most commonly used parts of RxDart to support nulls -- which I think are bad practice personally.

What are your thoughts?

lwasyl:
@brianegan I'm with you about the nulls. I'll close this PR for now and let's discuss in #170

brianegan:
After further discussion in #170, I think we should go ahead and revisit this PR. I don't love it, but the Stream API (StreamControllers) allows `null` values and we've decided long ago to go with the Dart conventions.

Not sure if you're still interesting in carrying this forward, but I'll provide a review and we can go from there :)

lib/src/subjects/behavior_subject.dart
```diff
@@ -59,7 +59,32 @@ class BehaviorSubject<T> extends Subject<T> implements ValueObservable<T> {
...
+        new Observable<T>.defer(
+            () => !wrapper.hasValue
+                ? controller.stream
+                : new Observable<T>(controller.stream)
+                    .startWith(wrapper.latestValue),
+            reusable: true),
+        wrapper);
+  }
+
+  factory BehaviorSubject({
```
brianegan: Is the logic in this factory correct? It appears to use the `() => wrapper.latestValue == null` logic instead of the `!wrapper.hasValue` logic.

brianegan: It looks like we might be missing a few tests around this constructor that would catch this bug!

brianegan:
@frankpepermans If you have a few minutes, would you mind reviewing this as well? We'll have to get it up to speed with the latest master, but overall I think we've had a lot of folks ask us to address this issue.

lwasyl:
@frankpepermans I think you can wait with review until I update the PR and make sure everything is implemented properly.

Also @brianegan please correct me if I'm wrong: the approach you settled on is that we'll break the API and add a `*seeded` version of constructors/factory methods for `BehaviorSubject` and `ValueObservable` that allows us to distinguish whether `null` seed value was passed. This way any `null` value will be treated exactly as a regularly emitted value. Is that right? So we'll have:
```dart
ValueConnectableObservable<T> publishValue()
ValueConnectableObservable<T> publishValueSeeded(T seedValue)

ValueObservable<T> shareValue()
ValueObservable<T> shareValueSeeded(T seedValue)

factory ValueConnectableObservable(Stream<T> source)
factory ValueConnectableObservable.seeded(T seedValue, Stream<T> source)

factory BehaviorSubject({onListen, onCancel, sync})
factory BehaviorSubject.seeded(T seedValue, {onListen, onCancel, sync})
```
Additionally we'll need `class ValueObservable<T>` to have new `bool get isNotEmpty` property, and `value` should now throw an exception when the observable hasn't been seeded and has no value, so
```dart
abstract class ValueObservable<T> implements Observable<T> {
  T get value; // throws StateException iff `isNotEmpty == false`
  
  bool get isNotEmpty;
}
```
Did I forget about something?

brianegan:
@lwasyl Thanks, that plan looks good to me!

Overall, the only part I wonder about is throwing an error if the value isn't set. Overall, I actually agree with doing that (fail fast!), but I wonder if end users will find it more frustrating than helpful?

E.g. they might currently assume the `value` returns null and they know how to handle that case, but in the new world that might mean they would see State Errors.

lib/src/observables/connectable_observable.dart
```diff
@@ -97,9 +98,20 @@ class ValueConnectableObservable<T> extends ConnectableObservable<T>
...
+  factory ValueConnectableObservable(Stream<T> source) =>
       ValueConnectableObservable<T>._(
-          source, BehaviorSubject<T>(seedValue: seedValue));
+        source,
+        BehaviorSubject<T>(),
+      );
+
+  factory ValueConnectableObservable.seeded(
+    Stream<T> source, {
+    @required T seedValue,
```
lwasyl: I've opted for `@required` seedValues since some factories/methods have more parameters

lib/src/observables/connectable_observable.dart
```diff
@@ -144,6 +156,9 @@ class ValueConnectableObservable<T> extends ConnectableObservable<T>
 
   @override
   T get value => _subject.value;
+
+  @override
+  bool get hasValue => _subject.hasValue;
```
lwasyl: I've opted for `hasValue`, since I found `isNotEmpty` harder to read. I know Brian mentioned `isNotEmpty` in the issue so I'm open to renaming it, but I like it without the negation

lib/src/subjects/behavior_subject.dart
```diff
@@ -59,15 +59,41 @@ class BehaviorSubject<T> extends Subject<T> implements ValueObservable<T> {
...
+    final wrapper = new _Wrapper<T>();
 
     return new BehaviorSubject<T>._(
         controller,
         new Observable<T>.defer(
-            () => wrapper.latestValue == null
-                ? controller.stream
-                : new Observable<T>(controller.stream)
-                    .startWith(wrapper.latestValue),
+            () => wrapper.hasValue
```
lwasyl: I've added a test that fails for the previous case where I mistakenly checked for last value's nullability

lib/src/observables/value_observable.dart
```diff
@@ -2,5 +2,9 @@ import 'package:rxdart/src/observables/observable.dart';
 
 /// An [Observable] that provides synchronous access to the last emitted item
 abstract class ValueObservable<T> implements Observable<T> {
+  /// Last emitted value, or null if there has been no emission yet
```
lwasyl: I've went for not throwing when there's no value, as per suggestion

lwasyl:
@brianegan @frankpepermans I've updated the PR. I've left some comments, and I'm open to suggestions. I think the PR now covers the null case appropriately

lib/src/observables/observable.dart
```diff
@@ -2502,8 +2538,43 @@ class Observable<T> extends Stream<T> {
...
+  /// // Subscribe again later. This will print 3 because it receives the last
+  /// // emitted value.
+  /// final subscription2 = observable.listen(print);
+  ///
+  /// // Stop emitting items from the source stream and close the underlying
+  /// // BehaviorSubject by cancelling all subscriptions.
+  /// subscription.cancel();
+  /// subscription2.cancel();
+  /// ```
+  ValueObservable<T> shareValueSeeded({@required T seedValue}) =>
```
brianegan: I'm interested what other folks think, but to me I kinda think `seedValue` should be a normal parameter in all of these cases. The current format will save folks who want to provide `null` seed values some code, but I still think that's more an edge case and we should optimize for folks who want to provide non-null values.

Therefore, this would read `shareValueSeeded(T seedValue)` and be called via `myObservable.shareValueSeeded(5)`. Repeating the `seedValue` in all of these constructors and methods feels a bit weird.

I kinda think all of the constructors should follow this path, but definitely open to discussion.

frankpepermans: Agree,

since we have an explicit constructor `shareValueSeeded`, then `shareValueSeeded(null)` is the same as `shareValueSeeded()`, in both cases we seed with null :)

The disambiguation we had before is gone.

I'm okay with this PR, yes many will need to refactor, but I'm sure it's not too big a deal.

lwasyl: Fixed. Hope I understood you correctly -- I removed all named parameters for seed value from all constructors and factory methods.

Also rebased the branch over the most recent changes with emitting error made by @frankpepermans, all tests pass so it's most likely okay. It does make sense to have more specific state in the wrapper, I like it.

Now I'm waiting till someone opens an issue to allow seeding the behavior subject with an error :D

brianegan: Lolol :P Thanks so much for your hard work on this one. Overall, I think it's looking good!

Issue #205, opened 2018-12-08
feature request pairwise


zoechi:
https://www.learnrxjs.io/operators/combination/pairwise.html

I think that would solve https://stackoverflow.com/questions/53675041/flutter-streams-how-to-access-previously-emitted-data-from-the-stream

I found it handy in some Angular TS situations.

frankpepermans:
At first glance, it looks a lot like bufferCount(2, 1).

I'll have a closer look later :)

frankpepermans:
see https://github.com/ReactiveX/rxdart/pull/210

Shantanu-Kotambkar:
@zoechi How would this solve my question on StackOverflow. https://stackoverflow.com/questions/53675041/flutter-streams-how-to-access-previously-emitted-data-from-the-stream

zoechi:
With each event you get the new value and the previous value.
If `[1, 2, 3, 4, 5]` are emitted one by one to the stream, pairwise would emit
`[1, null]`, `[2, 1]`, `[3, 2]`, `[4, 3]`, `[5, 4]`.
Is this not what you want?

Shantanu-Kotambkar:
Yes it is what i want, i just don't know how to use this. I was dealing with ReplaySubject, which isn't working for me. If you could answer my question on StackOverflow with a sample example, it would be just great!

zoechi:
It does not exist yet. `bufferCount` ;(nstead of pairwise) should do that when #210 is merged and an update published to pub.

zoechi:
A self-contained running example would probably be helpful to diagnose. 
The maintainers here know much more about that than me anyway.

frankpepermans:
For ReplaySubject, it always helps to check out the tests: https://github.com/ReactiveX/rxdart/blob/master/test/subject/replay_subject_test.dart

frankpepermans:
fixed in 0.20.0

PR #210, opened 2018-12-09
fixes buffer count and window count

frankpepermans:
This fix should properly handle `buffer` on `count` and `window` on `count`, when using the `startBufferEvery` parameter.

This *is* a breaking change:
- startBufferEvery should now work correctly
- when the parent `Stream` completes, we no longer add the remainder `buffer` as a final `event`:
i.e. if the `buffer` is set to group by 3 events, and the parent `Stream` outputs `1, 2, 3, 4`, then the only `event` on `buffer(3)` would be `[1, 2, 3]`. 
Before, we would also dispatch `[4]` as the remainder `onDone`.

### DIFF
--- a/lib/src/observables/observable.dart
+++ b/lib/src/observables/observable.dart
@@ -1198,44 +1198,56 @@ class Observable<T> extends Stream<T> {
               OnDataTransform<List<T>, List<T>> scheduleHandler) =>
           sampler(stream, bufferHandler, scheduleHandler)));
 
-  /// Creates an Observable where each item is a [List] containing the items
-  /// from the source sequence, in batches of [count].
-  ///
-  /// If [startBufferEvery] is provided, each group will start where the previous group
-  /// ended minus the [startBufferEvery] value.
+  /// Buffers a number of values from the source Observable by [count] then
+  /// emits the buffer and clears it, and starts a new buffer each
+  /// [startBufferEvery] values. If [startBufferEvery] is not provided or is
+  /// null, then new buffers are started immediately at the start of the source
+  /// and when each buffer closes and is emitted.
   ///
   /// ### Example
+  /// [count] is the maximum size of the buffer emitted
   ///
   ///     Observable.range(1, 4)
   ///       .bufferCount(2)
-  ///       .listen(print); // prints [1, 2], [3, 4]
+  ///       .listen(print); // prints [1, 2], [3, 4] done!
   ///
-  /// ### Example with startBufferEvery
+  /// ### Example
+  /// if [startBufferEvery] is 2, then a new buffer will be started
+  /// on every other value from the source. A new buffer is started at the
+  /// beginning of the source by default.
   ///
-  ///     Observable.range(1, 4).bufferCount(2, 1)
-  ///       .listen(print); // prints [1, 2], [2, 3], [3, 4], [4]
+  /// ### Example with [startBufferEvery]
+  ///
+  ///     Observable.range(1, 5).bufferCount(3, 2)
+  ///       .listen(print); // prints [1, 2, 3], [3, 4, 5], [5] done!
   Observable<List<T>> bufferCount(int count, [int startBufferEvery = 0]) =>
       transform(new BufferStreamTransformer<T>(
           onCount<T, List<T>>(count, startBufferEvery)));
 
   /// Deprecated: Please use [bufferCount]
   ///
-  /// Creates an Observable where each item is a [List] containing the items
-  /// from the source sequence, in batches of [count].
-  ///
-  /// If [startBufferEvery] is provided, each group will start where the previous group
-  /// ended minus the [startBufferEvery] value.
+  /// Buffers a number of values from the source Observable by [count] then
+  /// emits the buffer and clears it, and starts a new buffer each
+  /// [startBufferEvery] values. If [startBufferEvery] is not provided or is
+  /// null, then new buffers are started immediately at the start of the source
+  /// and when each buffer closes and is emitted.
   ///
   /// ### Example
-  ///
-  ///     Observable.range(1, 4).bufferWithCount(2)
-  ///       .listen(print); // prints [1, 2], [3, 4]
-  ///
-  /// ### Example with startBufferEvery
+  /// [count] is the maximum size of the buffer emitted
   ///
   ///     Observable.range(1, 4)
-  ///       .bufferWithCount(2, 1)
-  ///       .listen(print); // prints [1, 2], [2, 3], [3, 4], [4]
+  ///       .bufferCount(2)
+  ///       .listen(print); // prints [1, 2], [3, 4] done!
+  ///
+  /// ### Example
+  /// if [startBufferEvery] is 2, then a new buffer will be started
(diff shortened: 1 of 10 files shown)


frankpepermans:
argh, this broke a few `Window` tests because I've changed the `onDone` handling, will check...

frankpepermans:
fixed, window was calling `onDone` too soon, which also closed the `Sampler` too soon, causing the last event to not emit in some tests

lib/src/samplers/buffer_strategy.dart
```diff
@@ -196,31 +194,30 @@ class _OnCountSampler<T, S> extends StreamView<S> {
...
     var eventIndex = 0;
 
-    skip ??= 0;
+    startBufferEvery ??= 0;
 
     if (count == null) {
       throw new ArgumentError('count cannot be null');
-    } else if (skip > count) {
-      throw new ArgumentError('skip cannot be greater than count');
     }
```
VerTiGoEtrex: Not sure of your philosophy on bad input checking, but maybe add a check for `count < 0` -- currently negatives will throw deep inside the buffer handler code.

test/transformers/buffer_count_test.dart
```diff
@@ -35,15 +35,14 @@ void main() {
       expect(expectedOutput[count][0], result[0]);
       expect(expectedOutput[count][1], result[1]);
       count++;
-    }, count: 2));
+    }, count: expectedOutput.length));
   });
 
   test('rx.Observable.bufferCount.skip', () async {
```
VerTiGoEtrex: Maybe update the names of these tests to reflect the new behavior/param name. Or add a "todo" ;)

frankpepermans:
Good call :)

Added checks:
count needs to be `>= 1`
startBufferEvery needs to be `>= 0` (if `null`, then it defaults to 0)

renamed some tests, added the above checks in tests.

Finally, added a new op, `pairwise`, which is a shorthand for `bufferCount(2, 1)`, see https://github.com/ReactiveX/rxdart/issues/205

frankpepermans:
Yup that's why it's commented out,
I was following the docs where it stated:

"Collects values from the past as an array, and emits that array only when its size reaches bufferSize"

But I'm totally fine with allowing the final value in again.

Just need to doublecheck if pairwise has the same logic behind :)

frankpepermans:
@brianegan added the remaining buffer (if any) again when `onDone` triggers

lib/src/observables/observable.dart
```diff
@@ -1201,43 +1201,45 @@ class Observable<T> extends Stream<T> {
   /// Creates an Observable where each item is a [List] containing the items
   /// from the source sequence, in batches of [count].
   ///
-  /// If [skip] is provided, each group will start where the previous group
-  /// ended minus the [skip] value.
+  /// If [startBufferEvery] is provided, each group will start where the previous group
```
brianegan: I feel like this comment is just slightly difficult to understand / maybe a bit wrong with the new implementation? The RxJS docs use this description:

  > Interval at which to start a new buffer. For example if startBufferEvery is 2, then a new buffer will be started on every other value from the source. A new buffer is started at the beginning of the source by default.

lib/src/observables/observable.dart
```diff
@@ -1201,43 +1201,45 @@ class Observable<T> extends Stream<T> {
...
+          onCount<T, List<T>>(count, startBufferEvery)));
 
   /// Deprecated: Please use [bufferCount]
   ///
   /// Creates an Observable where each item is a [List] containing the items
   /// from the source sequence, in batches of [count].
   ///
-  /// If [skip] is provided, each group will start where the previous group
-  /// ended minus the [skip] value.
+  /// If [startBufferEvery] is provided, each group will start where the previous group
```
brianegan: Might need the same wording update as mentioned above

lib/src/observables/observable.dart
```diff
@@ -2291,23 +2304,24 @@ class Observable<T> extends Stream<T> {
...
+          new WindowStreamTransformer<T>(onCount(count, startBufferEvery)));
 
   /// Deprecated: Please use [windowCount]
   ///
   /// Creates an Observable where each item is a [Stream] containing the items
   /// from the source sequence, in batches of [count].
   ///
-  /// If [skip] is provided, each group will start where the previous group
-  /// ended minus the [skip] value.
+  /// If [startBufferEvery] is provided, each group will start where the previous group
```
brianegan: Wording

lib/src/samplers/buffer_strategy.dart
```diff
@@ -196,31 +194,35 @@ class _OnCountSampler<T, S> extends StreamView<S> {
...
+      [int startBufferEvery = 0]) {
     var eventIndex = 0;
 
-    skip ??= 0;
-
     if (count == null) {
       throw new ArgumentError('count cannot be null');
-    } else if (skip > count) {
-      throw new ArgumentError('skip cannot be greater than count');
+    } else if (count < 1) {
```
brianegan: Should we consider using `assert` here instead of throwing errors? That way these won't impact perf in production apps?

test/transformers/pairwise_test.dart
```diff
@@ -0,0 +1,48 @@
+import 'dart:async';
+
+import 'package:rxdart/rxdart.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('rx.Observable.pairwise', () async {
```
brianegan: I wouldn't adjust these tests, but in the future I think we should use Stream Matchers from the test package a bit more instead of these manual counts / expectAsync calls.

lib/src/samplers/buffer_strategy.dart
frankpepermans: Not sure, I suspect in most cases these args will be constant values, i.e. bufferCount(2, 1).

However should you write it this way: bufferCount(i, j) which could be anything, then your app might still fail in production.

It's only a null and an int compare check really.

Side note, I was building a benchmark test branch for all our code, so maybe we can revisit this later?

test/transformers/pairwise_test.dart
frankpepermans: Ah, lazy copy pasting of other tests, agreed ;)

lib/src/observables/observable.dart
frankpepermans: Yup will update

lib/src/samplers/buffer_strategy.dart
frankpepermans: So we keep the checks I assume?

brianegan: Yep, let's keep em for now!

frankpepermans:
Fixes `pairwise`, the diff to `bufferCount(2, 1)` here is, that `pairwise `does *not* emit the remaining buffer `onDone`.

e.g.
```
Observable.range(1, 4)
    .bufferCount(2, 1)
    .listen(print); // [1, 2], [2, 3], [3, 4], [4]
```

```
Observable.range(1, 4)
    .pairwise()
    .listen(print); // [1, 2], [2, 3], [3, 4]
```

Issue #213, opened 2018-12-10
Using asyncMap on a BehaviourSubject breaks broadcast behaviour


ferdkuh:
Using `asyncMap` on a `BehaviorSubject` seems to break the broadcast behavior.
When I add multiple subscriptions to the resulting stream I get a `Bad state: Stream has already been listened to` exception.  `asyncMap` on a `PublishSubject` or `StreamController.broadcast()` does not have this problem. Regular `map` or other transformations seem to work, i.e. keep the broadcast behavior.

The docs for `asyncMap` state that "The returned stream is a broadcast stream if this stream is.", so I would expect this to work.

Here's a snippet to reproduce the issue:

```
import 'package:rxdart/rxdart.dart';

void main() {

  Future<String> testMapAsync(String s) =>
      Future.delayed(Duration.zero, () => s.toUpperCase());

  String testMap(String s) => s.toUpperCase();

  var source = new BehaviorSubject<String>();

//  var mapped = source.map(testMap);             // this works
  var mapped = source.asyncMap(testMapAsync);     // results in BadState exception

  mapped.listen((x) => print("[1] $x"));
  mapped.listen((x) => print("[2] $x"));

  source.add("hello");

}
```

frankpepermans:
see https://github.com/ReactiveX/rxdart/pull/214

brianegan:
Fix merged! Will be shipped as part of RxDart `0.20`. Please keep an eye out for that!

PR #214, opened 2018-12-10
DeferStream isBroadcast getter was wrong

frankpepermans:
our `DeferStream` would always resolve `isBroadcast` to false, causing the issue seen in https://github.com/ReactiveX/rxdart/issues/213.

This fix treats the internal `_isReusable` as `isBroadcast`, which correctly propagates to the framework on subsequent transformations.

### DIFF
--- a/lib/src/streams/defer.dart
+++ b/lib/src/streams/defer.dart
@@ -20,6 +20,9 @@ class DeferStream<T> extends Stream<T> {
   final bool _isReusable;
   bool _isUsed = false;
 
+  @override
+  bool get isBroadcast => _isReusable;
+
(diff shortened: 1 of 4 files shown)


frankpepermans:
Think we're good on Subjects, the internal controllers are broadcast, and we expose their Streams as Observable.

Observable then exposes isBroadcast as _stream.isBroadcast.

The defer however is a special case where we didn't inherit

frankpepermans:
Added extra tests on `Publish` and `Replay` `Subjects` for the above

brianegan:
Fair point! You're right, and it's for the best since we were able to catch this particular bug.

PR #216, opened 2018-12-11
Adds repeat factory constructor

frankpepermans:
This PR will remove the old `repeat` operator, and replace it with a new `repeat` factory constructor instead.

The old `repeat` was very limited, it was basically just a simple `Transformer` which would repeat the last emitted event n-times.
Instead, the new constructor is based on the `RepeatStream`, which is very much alike `RetryStream` and allows us to replay a sequence n-times instead.

If folks want to use the old repeat, it can be very easily mimicked by:

```
Observable
    .just(1)
    .map((event) => [event, event, event])
    .expand((event) => event);
```


frankpepermans:
PS I've considered deprecating the old repeat operator, but I think keeping it would only cause confusion.
In other rx implementations, running repeat down the chain will replay the above sequence (which we cannot achieve with Dart Streams)

kentcb:
Thanks so much for this @frankpepermans - just getting around to trying it out. Was the `repeatIndex` parameter [left in](https://github.com/ReactiveX/rxdart/pull/216/files#diff-385e6a95078f4d968f00a82d418a6eefR619) by mistake? It doesn't appear to be used/make any sense.

kentcb:
Oh wait, nvm. I read the code as `repeatIndex` being a parameter to `repeat`, not a parameter to the factory. Now it makes sense!

kentcb:
And....it seems to work great! 🎉 Thanks again @frankpepermans!

frankpepermans:
Good to hear!

PR #217, opened 2018-12-12
adds mapTo operator

frankpepermans:
A very simple mapper, works by simply transforming each event into a constant value, see https://www.learnrxjs.io/operators/transformation/mapto.html

Thought it'd be useful to add, and makes the following example more terse:
```
Observable(element.click)
    .map((_) => 'clicked!');
```
becomes:
```
Observable(element.click)
    .mapTo('clicked!');
```

### DIFF
--- a/lib/src/observables/observable.dart
+++ b/lib/src/observables/observable.dart
@@ -1785,6 +1785,16 @@ class Observable<T> extends Stream<T> {
   Observable<S> map<S>(S convert(T event)) =>
       new Observable<S>(_stream.map(convert));
 
+  /// Emits the given constant value on the output Observable every time the source Observable emits a value.
+  ///
+  /// ### Example
+  ///
+  ///     Observable.fromIterable([1, 2, 3, 4])
+  ///       .mapTo(true)
+  ///       .listen(print); // prints true, true, true, true
+  Observable<S> mapTo<S>(S value) =>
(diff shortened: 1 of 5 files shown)


lib/src/observables/observable.dart
```diff
@@ -1785,6 +1785,16 @@ class Observable<T> extends Stream<T> {
...
 
+  /// Emits the given constant value on the output Observable every time the source Observable emits a value.
+  ///
+  /// ### Example
+  ///
+  ///     Observable.fromIterable([1, 2, 3, 4])
+  ///       .mapTo(true)
+  ///       .listen(print); // prints true, true, true, true
+  Observable<S> mapTo<S>(S value) =>
+      transform(new MapToStreamTransformer<T, S>(value));
```
brianegan: Do we need a Stream transformer for this? Couldn't we just do `return map((_) => value)`?

frankpepermans:
My only motivation was performance, no try catch involved, and no method execution.

But that's it really:)

brianegan:
Yah, it's a bit of a Catch22 for perf. 

It might help perf by foregoing a try / catch, but it also means folks who use this might end up sending a larger JS / Flutter app bundle to their users.

Overall, I'm about a "2 out of 10" in terms of how strongly I feel about removing the transformer, so if you like the current implementation and feel stronger, let's keep it :)

Issue #168, opened 2018-07-02
Add debounce operator with durationSelector


sir-boformer:
Right now rxdart only supports debouncing with a fixed duration, it would be great to have more control:

```dart
obs.debounceSelector((String event) {
  if(event == 'important') {
    return Observable.empty(); // forward immediately
  }
  else {
    return Observable.timer(null, Duration(seconds: 5)); // debounce 5 seconds
  }
});
````

In rxjs: http://reactivex.io/rxjs/class/es6/Observable.js~Observable.html#instance-method-debounce
(some more discussion: https://github.com/ReactiveX/rxjs/issues/493)

In RxJava: http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#debounce-io.reactivex.functions.Function-

frankpepermans:
Hi there! Yup that sure has some use cases, we do in fact have some samplers already, for buffer and window, see also https://github.com/ReactiveX/rxdart/blob/master/lib/src/samplers/buffer_strategy.dart

We could expand this to more ops as we go

frankpepermans:
see https://github.com/ReactiveX/rxdart/pull/249/

PR #220, opened 2018-12-13
Debounce selector

frankpepermans:
Small change for the `DebounceTransformer`, which now accepts a selector instead of a fixed `Duration`.

As requested in https://github.com/ReactiveX/rxdart/issues/168

On our `Observable`, we now have:
```
debounce(Duration duration);
debounceSelector(Duration durationSelector(T event));
```

### DIFF
--- a/lib/src/observables/observable.dart
+++ b/lib/src/observables/observable.dart
@@ -1326,15 +1326,51 @@ class Observable<T> extends Stream<T> {
   /// The Debounce operator filters out items emitted by the source Observable
   /// that are rapidly followed by another emitted item.
   ///
+  /// If [duration] is null, then the events are emitted immediately.
+  ///
   /// [Interactive marble diagram](http://rxmarbles.com/#debounce)
   ///
   /// ### Example
   ///
   ///     new Observable.range(1, 100)
-  ///       .debounce(new Duration(seconds: 1))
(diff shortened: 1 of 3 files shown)


brianegan:
One quick Q: Looking at the RxJS implementation, the durationSelector returns an Observable, rather than a `Duration`. Think we should do the same?

frankpepermans:
Oh then we'd need schedulers, like the rxjs timer. Maybe a different name for this one? I.e. debounceFor, debounceWith, ...

Or simply debounce with 2 parameters, so you choose to provide either a handler or a fixed Duration?

frankpepermans:
Looking at https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/debounce.ts, it accepts either an `Observable` or a `Promise` only. In fact all time-based implementations in rxjs do this, i.e. https://github.com/ReactiveX/rxjs/blob/master/src/internal/operators/throttle.ts as well.

So maybe we oughta consider changing this *everywhere* at some point.

Now for this PR, in the same context it might be better to just cancel it at this point, this PR would create another `debounceX` next to the old `debounce`, but we should be consistent then and do the same for `throttle` and `sample` for example.

If we then at some point change the API to accommodate to rxjs, we'd then have to drop these extra ops again.

So TL;DR we delete this PR and match rxjs better in the Future, for all timer ops?

Issue #227, opened 2018-12-27
BehaviorSubject disregards Exception and gives back last valid value when listend to. Intended?


JSANL:
In my flutter application I have a bloc. It not only sends data with a `BehaviorSubject`, but also exceptions (per `stream.addError()`), which I want to handle in the "front-end" by showing certain error messages to the user.

Let's say this is the bloc output:
- Output 1: valid event.
- Output 2: valid event.
- Output 3: Exception.

Now if I listen to the Stream after Output 3 was added then I will only receive Output 2 and know nothing about Output 3 (the Exception). In the case of my app Output 2 would reflect an old state and might confuse users.

My way of thinking was that the `BehaviorSubject` would emit the last Output to any new listener disregarding if it is a "normal" value or an `Exception`. So in this example Output 3 instead of Output 2.

Is this intended? If yes I think it should be stated in the documentation.

JSANL:
I tried solving it with a different `BehaviorSubject` emitting `Exception`s and having 2 `StreamBuilder`, first checking the `Exception` stream.
This is obviously not a really nice solution.

frankpepermans:
Fixed in master, soon on pub

JSANL:
Thank you very much! :)

PR #230, opened 2019-01-15
makes BehaviorSubject emit an Error, if the latest event was an Error…

frankpepermans:
…, https://github.com/ReactiveX/rxdart/issues/227

### DIFF
--- a/test/subject/behavior_subject_test.dart
+++ b/test/subject/behavior_subject_test.dart
@@ -35,6 +35,20 @@ void main() {
       await expectLater(subject, emits(3));
     });
 
+    test('emits errors to every subscriber', () async {
+      // ignore: close_sinks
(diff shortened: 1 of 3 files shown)


test/subject/behavior_subject_test.dart
```diff
@@ -35,6 +35,20 @@ void main() {
       await expectLater(subject, emits(3));
     });
 
+    test('emits errors to every subscriber', () async {
```
brianegan: Maybe add a check in here for `subject.value, isNull` as well and the opposite test for the `addError` -> `add` condition?

frankpepermans: sure! added meanwhile

frankpepermans:
@brianegan OK to merge for ya?

brianegan:
Yes sir!

Issue #229, opened 2019-01-14
doOnComplete called when there is an error after the first observable emits an item


Vanethos:
As an rxJava user (and I must preface this as stating I might be wrong), I think that `doOnComplete` was called only when the stream was finished without any error.

I ran some tests in RxDart and found the following:

```
Observable.just(1)
          .map((value) => throw Exception("message"))
          .doOnError((error, stacktrace) => print("Error"))
          .doOnDone(() => print("We are done!"))
          .listen(null);
```

prints
```
Error
We are done!
```

```
Observable.just(1)
          .flatMap((value) => throw Exception("message"))
          .doOnError((error, stacktrace) => print("Error"))
          .doOnDone(() => print("We are done!"))
          .listen(null);
```

prints
```
Error
We are done!
```

Now, shouldn't the `doOnDone` (which is the same as `doOnComplete` as per documentation) be called _only_ if there is no error in the stream?

brianegan:
This is indeed a difference between Dart streams and Rx. Errors are, by default, not terminating events.

You can make them terminating events by passing through: `.listen(null, cancelOnError: true)`. In that case, you should not see the "onDone."

Vanethos:
oh thanks! Didn't catch that, was about to create a transformer to change `cancelOnError`, thanks!

PR #231, opened 2019-01-19
Added `onErrorCancel` transformer to the stream

Vanethos:
… #229


brianegan:
Hey there -- sorry, somehow missed this PR! Overall, I'm not quite sure I understand the intention of this change.

It seems like this forces Streams to be cancelled on error, but I worry that would have a really weird API in the end. E.g. even if the user sets `cancelOnError: true` in the `listen` method, it won't be respected.

Overall, I sympathize with the goal, but I don't know if if this type of feature makes as much sense in Dart.

Vanethos:
Hey Brian!

Well, I have to say I agree with you, this was something I made before I found out about `cancelOnError` and, after learning about it, I still sent a pull request just for a reason: I do like to control the stream in the chain instead of using the listen method (so using `doOnError` instead of the listen's `onError` for instance), and I thought that this could be useful to someone.

If it isn't, welp, at least I tried :D 

Thanks for the feedback!

brianegan:
Sounds good, and totally makes sense! I think we should go ahead and close this out for now since Dart seems to leave this determination up to the subscriber instead of the producer!

Thanks for contributing :)

Issue #235, opened 2019-02-01
Bring `groupBy` operator back.


yohom:
As https://github.com/ReactiveX/rxdart/issues/127#issuecomment-373939673 said. Thanks!

frankpepermans:
groupBy is now available in master, soon on pub

PR #238, opened 2019-02-05
adds groupBy

frankpepermans:
Adds groupBy (see https://github.com/ReactiveX/rxdart/issues/235)

This operator takes a `Function`, this method returns a `key` based on the current `event`.
With this `key`, a new `Observable` is created (a `GroupByObservable` which has an extra `key` property).

All events that return the same `key`, are added to the same `GroupByObservable` with that same `key`.


svsobh:
Thanks a lot for this @frankpepermans !

Dear @brianegan ,

May you please review and merge these changes as soon as you get some time:) It will be really helpful!

Have a great day both of you!

PR #245, opened 2019-03-12
Updated documentation and example for combineLatest

DarshanGowda0:
Updated the documentation and example of combineLatest, which is [link](https://github.com/ReactiveX/rxdart/blob/b0f08ac8057b9ba966b6afb3c2b3c007094d6b53/lib/src/observables/observable.dart#L96) same as combineLatestList at the moment.


brianegan:
Excellent, thanks!

Issue #175, opened 2018-08-07
Throttle: configure leading/trailing events


greglittlefield-wf:
It would be very useful to be able to configure `ThrottleStreamTransformer` to emit trailing events, as is possible in the JS implementation: https://rxjs-dev.firebaseapp.com/api/operators/throttle

brianegan:
Thanks! Good call -- we can definitely add this.

frankpepermans:
see https://github.com/ReactiveX/rxdart/pull/249/

PR #246, opened 2019-03-12
refactors all backpressure transformers

frankpepermans:
The current backpressure transformers were a bit inconsistent:
- `buffer` and `window` used a `Sampler` strategy
- `debounce` and `throttle` took a `Duration`, while `sample` took a `Stream`

This refactor is an attempt to simplify things.
All backpressure transformers now extend the base `BackpressureStreamTransformer` class.

`buffer`, `window`, `debounce`, `throttle`, `pairwise` and `sample` are now just other configurations of this `BackpressureStreamTransformer`.

I've removed `bufferFuture` and `windowFuture` (these can be easily mimicked).

The other ones are now exposed as:
- `debounce(Stream window(T event))` (allows to return a new window on a next debounced value)
- `debounceTime(Duration duration)` is the old one
- `throttle(Stream window(T event))`
- `throttleTime(Duration duration)`
- `sample(Stream sampleStream)`
- `sampleTime(Duration duration)`

@brianegan Currently coverage is broken ;-/ Dunno yet how to fix it, any ideas welcome!


frankpepermans:
https://github.com/dart-lang/coverage/issues/249

brianegan:
Wow, amazing work @frankpepermans!!! I'll review this a bit later today.

One Q: Does this new abstraction allow us to support leading and trailing events for `throttle`? https://github.com/ReactiveX/rxdart/issues/175

brianegan:
Hrm, the test coverage bit is odd. I just merged another PR and master is still passing.

Do you know if there might be a new test somewhere in here that's hanging, similar to what we experienced in the past?

frankpepermans:
@brianegan depends, 
According to rxjava, it's actually the same as sample, see https://github.com/ReactiveX/RxJava/blob/v1.0.16/src/main/java/rx/Observable.java#L8819-L8821

But there is a subtle difference between `throttle` and `sample`:
- `sample` takes another `Stream`, and emits whenever that `Stream` emits
- `throttle` waits for an event, then opens a window, when that window completes, it emits the first event emitted within that window.

So `sample` does not reset on events, where the window of `throttle` does.

It's better explained with marbles maybe:
https://rxmarbles.com/#sample
https://rxmarbles.com/#throttle

In rxjs land, they are indeed different:, see https://stackoverflow.com/questions/48053788/difference-between-sample-and-throttle-in-rxjs

TL;DR, yes easy to add, didn't do it yet because of the above explanation.
I'd say yes, add it though.

frankpepermans:
My PR runs on Dart 2.2.0 stable, I've changed the sdk as it was still 2.0.0-dev before.

This change breaks it

PR #248, opened 2019-03-13
Purposely add a failing test to ensure we don't need to run tests twice

brianegan:
In the new travis script, I stopped running the tests twice since that's exactly what the coverage team does in their travis setup. I want to ensure that's safe -- if we get a test failure, travis should fail.

Ensure the new travis script is working as expected when a test fails.


brianegan:
Shouldn't have passed. We'll fix the test script.

PR #251, opened 2019-03-14
adds forkJoin Stream

frankpepermans:
For "completion"'s sake :P

Based on `combineLatest`, so that gives:
- `forkJoin`
- `forkJoinList`
- `forkJoin2` ... `forkJoin9`

Like `combineLatest`, you can provide a combiner `Function`. (which are Typed, if using any of the forkJoin-n ctrs)

### DIFF
--- a/lib/src/streams/fork_join.dart
+++ b/lib/src/streams/fork_join.dart
@@ -1,0 +1,294 @@
+import 'dart:async';
+
+/// This operator is best used when you have a group of observables
+/// and only care about the final emitted value of each.
+/// One common use case for this is if you wish to issue multiple
+/// requests on page load (or some other event)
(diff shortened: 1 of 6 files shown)


lib/src/streams/fork_join.dart
```diff
@@ -0,0 +1,294 @@
...
+///       Stream.fromIterable(["a"]),
+///       Stream.fromIterable(["b"]),
+///       Stream.fromIterable(["C", "D"])])
+///     .listen(print); //prints ['a', 'b', 'D']
+///
+/// ### Example with combiner
+///
+/// If you wish to combine the list of values into a new object before you
+///
+///     CombineLatestStream(
```
brianegan: Copy Pasta?

PR #253, opened 2019-03-18
makes zip use less resources

frankpepermans:
I've reworked zip, in the old version, each inner `subscription` had its own `List` to buffer events,
then when each of those Lists had at least one `event`, we iterate over the Lists, remove the first event in each, and bundle those as arguments for the `zipper`.

This could result in a resource hog if we, for example, zip 2 Streams where the first one emits thousands of events continuously, while the other one only submits on long intervals.

The rework now immediately `pauses` a subscription after it emits an event, 
when all subscriptions are paused, then they have all emitted one event, so we now bundle these and pass them to the `zipper`.

Then, we simply `resume` them all and the process repeats.

### DIFF
--- a/lib/src/streams/zip.dart
+++ b/lib/src/streams/zip.dart
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:collection';
 
 /// Merges the specified streams into one observable sequence using the given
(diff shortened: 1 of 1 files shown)


brianegan:
Will this work for broadcast streams?

PR #264, opened 2019-04-05
Docs: IntervalStreamTransformer accepts Duration

f3ath:
The doc block is out of sync with the actual code


brianegan:
Oh, good call -- thanks!

PR #262, opened 2019-04-02
Docs: Fix ZipStream example and usage of Stream.fromIterable()

apaatsio:
1. Change examples using `Stream.fromIterable(a, b)` to `Stream.fromIterable([a, b])`.
2. ZipStream constructor accepts a zipper function `R zipper(List<T> values)` but the example was using `(a, b) => a + b` which doesn't match the function signature. Fixed and updated the example.
3. Added an example of `ZipStream.zip2` making the docs more consistent with e.g. `CombineLatestStream` and `ForkJoinStream`.

PR #269, opened 2019-04-16
fix: compilation errors in examples applying `dartanalyzer`

exKAZUu:
c.f. https://github.com/ReactiveX/rxdart/pull/262#issuecomment-482065922

This PR applies `dartanalyzer` to the dart files in `example` using another `analysis_options.yaml`.
The `analysis_options.yaml` ignores existing semantical errors but detects `unnecessary_new` errors.
I believe this approach is useful to maintain the dart files in `example`.

### DIFF
--- a/example/analysis_options.yaml
+++ b/example/analysis_options.yaml
@@ -1,7 +1,6 @@
 analyzer:
   errors:
     argument_type_not_assignable: ignore
-    deprecated_member_use: ignore
     extends_non_class: ignore
     implements_non_class: ignore
     non_type_as_type_argument: ignore
--- a/example/analysis_options.yaml
+++ b/example/analysis_options.yaml
@@ -1,3 +1,4 @@
+include: package:pedantic/analysis_options.yaml
 analyzer:
   errors:
     argument_type_not_assignable: ignore
--- a/example/analysis_options.yaml
+++ b/example/analysis_options.yaml
@@ -1,0 +1,20 @@
+analyzer:
+  errors:
+    argument_type_not_assignable: ignore
+    deprecated_member_use: ignore
+    extends_non_class: ignore
+    implements_non_class: ignore
+    non_type_as_type_argument: ignore
+    override_on_non_overriding_method: ignore
+    undefined_annotation: ignore
+    undefined_class: ignore
+    undefined_function: ignore
+    undefined_identifier: ignore
+    undefined_method: ignore
+    undefined_named_parameter: ignore
+    undefined_super_method: ignore
+    uri_does_not_exist: ignore
+    type_test_with_undefined_name: ignore
+linter:
+  rules:
+    - unnecessary_new
--- a/example/flutter/github_search/lib/search_bloc.dart
+++ b/example/flutter/github_search/lib/search_bloc.dart
@@ -15,7 +15,7 @@ class SearchBloc {
         // If the text has not changed, do not perform a new search
         .distinct()
         // Wait for the user to stop typing for 250ms before running a search
-        .debounce(const Duration(milliseconds: 250))
+        .debounceTime(const Duration(milliseconds: 250))
         // Call the Github api with the given search term and convert it to a
         // State. If another search term is entered, flatMapLatest will ensure
         // the previous search is discarded so we don't deliver stale results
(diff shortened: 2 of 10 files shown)


example/analysis_options.yaml
```diff
@@ -0,0 +1,20 @@
...
+    undefined_function: ignore
+    undefined_identifier: ignore
+    undefined_method: ignore
+    undefined_named_parameter: ignore
+    undefined_super_method: ignore
+    uri_does_not_exist: ignore
+    type_test_with_undefined_name: ignore
+linter:
+  rules:
+    - unnecessary_new
```
exKAZUu: For now, I just put only `unnecessary_new`. If this direction is okay, then we should add more options here.

example/analysis_options.yaml
```diff
@@ -0,0 +1,20 @@
+analyzer:
+  errors:
+    argument_type_not_assignable: ignore
+    deprecated_member_use: ignore
```
exKAZUu: I think we should enable this option after merging this PR.

exKAZUu: I've re-enabled `deprecated_member_use`.

example/web/search_github/search_github.dart
```diff
@@ -23,7 +23,7 @@ void main() {
       .where((term) => term.isNotEmpty)
       // Use debounce() to prevent calling the server on fast following
       // keystrokes
-      .debounce(const Duration(milliseconds: 250))
+      .debounceTime(const Duration(milliseconds: 250))
```
exKAZUu: The previous code caused a compilation error, so I've fixed it.

example/flutter/github_search/lib/search_bloc.dart
```diff
@@ -15,7 +15,7 @@ class SearchBloc {
         // If the text has not changed, do not perform a new search
         .distinct()
         // Wait for the user to stop typing for 250ms before running a search
-        .debounce(const Duration(milliseconds: 250))
+        .debounceTime(const Duration(milliseconds: 250))
```
exKAZUu: The previous code caused a compilation error, so I've fixed it.

example/flutter/github_search/test/search_bloc_test.dart
```diff
@@ -99,12 +99,12 @@ void main() {
   });
 }
 
-const noTerm = isInstanceOf<SearchNoTerm>();
+const noTerm = TypeMatcher<SearchNoTerm>();
```
exKAZUu: `isInstanceOf` is deprecated and we should use `TypeMatcher`.

brianegan:
Hey, thanks so much! Sorry about the delay, back from vacay. Reviewing now

example/analysis_options.yaml
```diff
@@ -0,0 +1,19 @@
+analyzer:
```
brianegan: Would it make sense to use the `pendantic` package's rules here and just stick to those, rather than making up our own set?

example/analysis_options.yaml
brianegan: Oops, see pedantic package comment above!

exKAZUu:
Though `pedantic` doesn't cause new errors, `pedantic` doesn't include `unnecessary_new`.

How about using `include: ../analysis_options.yaml` and disabling rules which cause lint errors instead of `pedantic`. This approach yields the following `analysis_options.yaml`.

```
include: ../analysis_options.yaml

analyzer:
  strong-mode:
    implicit-casts: true
    implicit-dynamic: true
  errors:
    argument_type_not_assignable: ignore
    extends_non_class: ignore
    implements_non_class: ignore
    non_type_as_type_argument: ignore
    override_on_non_overriding_method: ignore
    undefined_annotation: ignore
    undefined_class: ignore
    undefined_function: ignore
    undefined_identifier: ignore
    undefined_method: ignore
    undefined_named_parameter: ignore
    undefined_super_method: ignore
    uri_does_not_exist: ignore
    type_test_with_undefined_name: ignore
linter:
  rules:
    close_sinks: false
    type_annotate_public_apis: false
```

exKAZUu:
Hmm, but `unnecessary_new` will be added in `pedantic` soon (c.f. https://github.com/dart-lang/pedantic/issues/18)

exKAZUu:
Okay, I've just added `pedantic`.

PR #266, opened 2019-04-10
fix: deal with NoSuchMethodError exception

exKAZUu:
I'm facing the following error. This PR tries fixing this error.

```
[VERBOSE-2:ui_dart_state.cc(148)] Unhandled Exception: NoSuchMethodError: The method 'cancel' was called on null.
Receiver: null
Tried calling: cancel()
#0      Object.noSuchMethod (dart:core/runtime/libobject_patch.dart:50:5)
#1      SwitchMapStreamTransformer._buildTransformer.<anonymous closure>.<anonymous closure> (package:rxdart/src/transformers/switch_map.dart:76:55)
<asynchronous suspension>
#2      _StreamController._recordCancel (dart:async/stream_controller.dart:713:20)
#3      _ControllerSubscription._onCancel (dart:async/stream_controller.dart:841:24)
#4      _BufferingStreamSubscription._cancel (dart:async/stream_impl.dart:242:21)
#5      _BufferingStreamSubscription.cancel (dart:async/stream_impl.dart:194:7)
#6      _StreamBuilderBaseState._unsubscribe (package:flutter/src/widgets/async.dart:154:21)
#7      _StreamBuilderBaseState.didUpdateWidget (package:flutter/src/widgets/async.dart:117:9)
#8      StatefulElement.update (package:flutter/src/widgets/framework.dart:3879:58)
#9      Ele<…>
```

PR #284, opened 2019-05-07
fix: missing null check

larssn:
I noticed that if something threw an Exception inside a switchMap, line 76 would crash with calling `cancel()` on a null. I looked at the file, and every other cancel of `otherSubscription`, already has the null check, so it just appears to be an oversight.

Do note that if `otherSubscription` is null, the line:
`if (hasMainEvent) await otherSubscription?.cancel();`
effectively becomes:
`if (hasMainEvent) await null;`

FYI `await` supports nulls, so there's no need to check for it.


larssn:
Apologies, appears to identical to #266

Issue #289, opened 2019-05-18
BehaviorSubject & switchmap crash


modulovalue:
[Stackblitz](https://stackblitz.com/edit/rxjs-xtkbgy)
Typescript RxJS behavior:
```ts
console.clear();
import { BehaviorSubject } from 'rxjs';
import { switchMap } from 'rxjs/operators';

const a = new BehaviorSubject(1);
const a1 = new BehaviorSubject(2);
const a2 = new BehaviorSubject(3);
const sut = a.pipe(
  switchMap(val => (val == 1 ? a1 : a2)),
);

a1.subscribe(console.log);
sut.subscribe(console.log);

a1.complete();
a.next(2);
a.complete();
a2.next(10);
```
output:
```
2
2
3
10
```
in RxDart
```dart
    final a = BehaviorSubject.seeded(1);
    final a1 = BehaviorSubject.seeded(2);
    final a2 = BehaviorSubject.seeded(3);

    final sut = a.switchMap((int val) => val == 1 ? a1 : a2);

    a1.listen(print);
    sut.listen(print);

    await a1.close();
    a.add(2);
    await a.close();
    a2.add(10);
```

output:
```
2
2
3
dart:async                                              _BoundSubscriptionStream.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/streams/defer.dart 36:18             DeferStream.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/transformers/switch_map.dart 47:55   SwitchMapStreamTransformer._buildTransformer.<fn>.<fn>.<fn>
===== asynchronous gap ===========================
dart:async                                              _BoundSubscriptionStream.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/streams/defer.dart 36:18             DeferStream.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/observables/observable.dart 1980:20  Observable.listen
package:rxdart/src/transformers/switch_map.dart 40:38   SwitchMapStreamTransformer._buildTransformer.<fn>.<fn>
dart:async                                              _StreamImpl.listen
package:rxdart/src/transformers/switch_map.dart 79:32   SwitchMapStreamTransformer._buildTransformer.<fn>
dart:async                                              _BoundSubscriptionStream.listen
[...]

PR #290, opened 2019-05-18
Potential fix for #289

modulovalue:


lib/src/transformers/switch_map.dart
```diff
@@ -58,22 +58,22 @@ class SwitchMapStreamTransformer<T, S> extends StreamTransformerBase<T, S> {
                 onDone: () {
                   leftClosed = true;
 
-                  if (rightClosed || !hasMainEvent) controller.close();
+                  if (rightClosed && !hasMainEvent) controller.close();
```
frankpepermans: If switchMap then received at least one event, and afterwards left and right close, then the switchMap controller would never fire onDone.

For example:
```
final a = BehaviorSubject.seeded(1);
        final a1 = BehaviorSubject.seeded(2);
        final a2 = BehaviorSubject.seeded(3);

        final sut = a.switchMap((int val) => val == 1 ? a1 : a2);
        expect(sut, emitsInOrder(<dynamic>[2, emitsDone]));

        await a1.close();
        await a2.close();
        await a.close();
        
        // emits 2
        // never emits 'done'
```

modulovalue: I've added your example as a test, kept my other new failing test and revoked the change to switchMap.

Issue #291, opened 2019-05-18
Observable.combineLatestList<int>([]) doesn't finish


modulovalue:
The following Observable 
`Observable.combineLatestList<int>([])`
never ends.

I expected:
```dart
final combinedEmptyList = Observable.combineLatestList<int>([]);
expect(combinedEmptyList, emitsInOrder(<dynamic>[emitsDone])); /// timeout
```

PR #292, opened 2019-05-18
Potential fix for #291

modulovalue:


lib/src/streams/combine_latest.dart
```diff
@@ -264,7 +264,12 @@ class CombineLatestStream<T, R> extends StreamView<R> {
 
     controller = StreamController<R>(
       sync: true,
-      onListen: () {
+      onListen: () async {
+        if (len == 0) {
```
frankpepermans: Good catch!
However I wonder if we should't just assert here that the list cannot be empty? @brianegan

PR #303, opened 2019-07-19
fix: check type of onError in DoStreamTransformer

hoc081098:
Before:
```dart
new Observable.error(new Exception())
    .doOnError((error, stacktrace) => print("oh no"))
    .listen(null); // prints "Oh no"
```
After: (can omit stacktrace parameter)
```dart
new Observable.error(new Exception())
    .doOnError((error) => print("oh no"))
    .listen(null); // prints "Oh no"
```


brianegan:
Hm, I'm a little mixed on this one. This almost turns the callback passed into this method `dynamic` with internal Type checking. I'd almost prefer we just specify that `onError` is given both and a signature that enforces this pattern so folks have proper IDE autocompletion, rather than being able to pass in any random `Function`.

What are your thoughts? @frankpepermans?

hoc081098:
I think it suits Dart Stream methods (`onError` parameter in [Stream.handleError](https://api.dartlang.org/stable/2.4.1/dart-async/Stream/handleError.html), [Stream.listen](https://api.dartlang.org/stable/2.4.1/dart-async/Stream/listen.html)). 
Just add document to make it clear

frankpepermans:
The Dart `Stream` `onError` handler is of `Type` `Function`, behind the scenes, this invoker is called to execute that Function: https://github.com/dart-lang/sdk/blob/master/sdk/lib/async/async_error.dart

I do prefer having it typed to `void onError(dynamic e, [StackTrace s])` myself however :)

We could call the `Function` found inside async_error, but unfortunately it is private [facepalm]

PR #310, opened 2019-08-05
Streams/from callable

frankpepermans:
Adds the fromCallable ctr.

This constructor allows you to create a new Observable from the result of a Function.

Each time you listen to the Observable, the Function should execute again.

The result is of type FutureOr<T>, the Observable will await a Future, if needed, and always yield a Stream of T.

### DIFF
--- a/lib/src/observables/observable.dart
+++ b/lib/src/observables/observable.dart
@@ -4,6 +4,7 @@ import 'package:rxdart/futures.dart';
 import 'package:rxdart/src/observables/connectable_observable.dart';
 import 'package:rxdart/src/observables/replay_observable.dart';
 import 'package:rxdart/src/observables/value_observable.dart';
+import 'package:rxdart/src/streams/from_callable.dart';
 import 'package:rxdart/streams.dart';
 import 'package:rxdart/transformers.dart';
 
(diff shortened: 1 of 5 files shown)


brianegan:
I dig it.

lib/src/observables/observable.dart
```diff
@@ -729,7 +730,24 @@ class Observable<T> extends Stream<T> {
...
-  /// Creates an Observable from the future.
+  /// Creates an [Observable] that, when subscribed upon,
+  /// invokes a function you specify and then emits the value returned from that function.
+  ///
+  /// The function expects a return type of [FutureOr]
+  /// - if you return a [Future], it will be auto-awaited, so there is no need
+  ///   to do [asyncMap] immediately.
+  /// - if you return a non-future, then that value is simply returned.
+  ///
+  /// [Observable]s created by fromCallable are always broadcast [Observable]s.
```
brianegan: This might be a bit mislead, and is the only thing I'm wondering about... If you subscribe to the Stream twice, it will execute the `callable` again and emit the values, but it won't emit the same data to all subscribers.

It's really more of a cold observable, which doesn't fit nicely into the Single-Subscription / Broadcast dynamic that cleanly :/

frankpepermans: Hehe indeed :)

Should we document this better?

i.e. the alternative would be something like: `Observable.fromFuture(callable())`, where all subscribers get the exact same data.

hoc081098:
Simpler version 

```dart
Rx.defer(() {
    final v = callable();
    return v is Future
        ? Stream.fromFutute(v)
        : Stream.value(v);
});
```

_Sent from my Redmi  7A using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #314, opened 2019-08-12
replaces of_type with where_type

frankpepermans:
Replaces the old ofType operator with whereType.
* This is consistent with List.whereType
* The old TypeToken is removed in favor of the following syntax:
`Observable(stream).whereType<T>()`


brianegan:
Hey, I was wondering about this one myself! Two bits of feedback:

  1. I think they might add this to the default Stream type, so I'm not 100% sure if it's a great idea for us to fill this in on our side? https://github.com/dart-lang/sdk/issues/34050
  2. If we want to make this change, I would say let's @deprecate `ofType` to give folks a couple of months to adjust, rather than removing that method, since I think it might be quite commonly used.

Other than that, the code generally looks good, but just a bigger question about whether we want to add something here that might conflict with a Future version of Dart?

frankpepermans:
I recall that this was effectively added at one point, in the Dart 2.0.0-dev builds, but later removed again for some reason.

I dunno, any operator we have just might be supported in the SDK somewhere down the road, whenever that happens, we will need to just update asap so that at least the latest release is compatible with the new SDK?

+1 on having deprecated of course

brianegan:
Yah, definitely agree any method could be added that conflicts (as has happened in the past, haha). Also hard to know when they would add it if it hasn't made it into Dart 2 yet...

I'd say let's go for it. Since Dart 2 allows us to properly infer types from generic methods, we no longer need that ugly `TypeToken`, which is a big win for readability.

lib/src/observables/observable.dart
```diff
@@ -2076,37 +2077,6 @@ class Observable<T> extends Stream<T> {
...
-  ///     new Observable.fromIterable([1, "hi"])
-  ///       .ofType(kString)
-  ///       .listen(print); // prints "hi"
-  ///
-  /// If you'd like to create your own shortcuts like the example above,
-  /// simply create a constant:
-  ///
-  ///     const TypeToken<Map<Int, String>> kMapIntString =
-  ///       const TypeToken<Map<Int, String>>();
-  Observable<S> ofType<S>(TypeToken<S> typeToken) =>
```
brianegan: `@deprecated`

lib/src/observables/observable.dart
```diff
@@ -2482,6 +2452,25 @@ class Observable<T> extends Stream<T> {
...
+  /// Events that do not match [T] are filtered out, the resulting
+  /// [Observable] will be of Type [T].
+  ///
+  /// ### Example
+  ///
+  ///     Observable.fromIterable([1, 'two', 3, 'four'])
+  ///       .whereType<int>()
+  ///       .listen(print); // prints 1, 3
+  ///
+  /// ## as opposed to:
```
brianegan: This will create an h2 heading that's larger than the h3 "Example" heading above... Maybe Leave out any heading here or use an h4?

lib/src/utils/type_token.dart
```diff
@@ -1,42 +0,0 @@
...
-/// checks and castings within an ofTypeObservable itself. Therefore, this class
-/// was introduced to capture the type of class you'd like `ofType` to filter
-/// down to, or `cast` to cast to.
-///
-/// ### Example
-///
-///     new Stream.fromIterable([1, "hi"])
-///       .ofType(new TypeToken<String>)
-///       .listen(print); // prints "hi"
-class TypeToken<S> {
```
brianegan: `@deprecated`

test/transformers/where_type_test.dart
```diff
@@ -23,22 +22,29 @@ Stream<Object> _getStream() {
 }
 
 void main() {
-  test('rx.Observable.ofType', () async {
```
brianegan: Not a lot of fun, but probs good to just duplicate these tests into a new file until we fully deprecate

frankpepermans:
@brianegan ofType is added again, as deprecated, tests too

brianegan:
LGTM! Nice one :)

PR #319, opened 2019-08-20
Update sample to latest Flutter stable

### LINKED ISSUE #317: Github Search example not working
Hi, I'm getting this issue when trying to run github_search app in `example` folder: 
```
[VERBOSE-2:ui_dart_state.cc(148)] Unhandled Exception: type 'List<String>' is not a subtype of type 'GithubApi' of 'api'
```

Why do we have `api` argument as Github Api as parameter? I though that main function only accept list of String arguments.

```dart
void main(GithubApi api) {
  runApp(SearchApp(api: api));
}
```

https://github.com/ReactiveX/rxdart/blob/master/example/flutter/github_search/lib/main.dart#L5

brianegan:
Fixes #317

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ dart:
   - dev
   - stable
 with_content_shell: false
-before_install:
-  - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
 script:
   - set -e
   - sh tool/check_format.sh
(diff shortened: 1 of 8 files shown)


brianegan:
Note: Updated travis.yml to remove xvfb since it was causing failures and isn't required for our test suite

brianegan:
CodeCov can't report the diff because master is also failing due to the xvfb thing, so once we merge this we should be good to go again.

PR #321, opened 2019-08-21
Fix seeded side of BehaviorSubject

frankpepermans:
The seeded factory `defer` method was not updated in the same way the non-seeded `defer` method was, since these are basically the same methods, I've created a static `Function` which both now call.


frankpepermans:
done!

brianegan:
Overall, LGTM, but wondering if we should we add a test case for this fix to prevent future regressions?

frankpepermans:
@brianegan yes, we do only test for non-seeded BehaviorSubjects, I've duped those which expect a value after a listen (i.e. value or Error or ...)

The tests below are fine as they were, I think, since the factory ctr plays no difference there

PR #325, opened 2019-09-03
Remove tween doc link

mhadaily:
it seems it was removed while back but it was still in the readme and linked to 404 pages.


brianegan:
Good call, thanks!

PR #330, opened 2019-09-17
switchIfEmpty: close stream when errored

bytepoets-mzi:
Hi,

the behavior of switchIfEmpty doesn't handle exceptions in the stream: when an exception is thrown, you receive an error but you **also** get the fallback value. I'm not too much into ReactiveX, but the following code works in RxJava:

```kotlin
val observer = TestObserver<Int>()
Observable
        .error<Int>(RuntimeException())
        .switchIfEmpty(Observable.just(1))
        .subscribe(observer)

observer.assertNoValues()
observer.assertError(RuntimeException::class.java)
```

But it doesn't work in rxdart.

PS: Is there an `assert(false)` in Dart? `expect(true, false)` does the job, but …

### DIFF
--- a/lib/src/transformers/switch_if_empty.dart
+++ b/lib/src/transformers/switch_if_empty.dart
@@ -60,7 +60,10 @@ class SwitchIfEmptyStreamTransformer<T> extends StreamTransformerBase<T, T> {
                   hasEvent = true;
                   controller.add(value);
                 },
-                onError: controller.addError,
+                onError: (Object err, [StackTrace stacktrace]) {
+                  hasEvent = true;
+                  controller.addError(err, stacktrace);
+                },
                 onDone: () {
                   if (hasEvent) {
                     controller.close();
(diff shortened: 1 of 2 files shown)


frankpepermans:
Hi there,

In Dart `Streams`, you specify whether or not your `subscription` terminates when an `Error` occurs,
by default, the `Stream` just keeps going *after* an `Error` is thrown:

So if you do not want the fallback to take place, provide `cancelOnError` true when using `listen`.

brianegan:
It's actually a really interesting question -- what does "Empty" mean in terms of Dart Streams. If an error is emitted, does that mean the Stream is no longer "empty"?

From one perspective -- yes! Because the stream emitted SOMETHING, just not a data event.

From another perspective -- no! Because it hasn't emitted any data events.

IMO, I like this change. I could see that emitting an error event does not mean the Stream is completely empty and that is has some kind of event. Therefore, I'm 👍 on making this change.

What do you think, Frank?

frankpepermans:
Tough one!

Been thinking about it, and could't decide pro or con.
It doesn't help that in Dart, a Stream just can go on after an Error (or not, if you flag it to cancel).

How about we instead provide switchIfEmpty(bool ignoreErrors)?

PR #332, opened 2019-09-19
export StreamSubscription

lukepighetti:
Hi there, instead of making an issue I just made a lite-PR. I'm always importing StreamSubscription to manage subscriptions and I'm wondering if we should just export it in rxdart. Thoughts?

### DIFF
--- a/lib/rxdart.dart
+++ b/lib/rxdart.dart
@@ -1,5 +1,7 @@
 library rx;
 
+export 'dart:async' show StreamSubscription;
+
 export 'package:rxdart/futures.dart';
 export 'package:rxdart/src/observables/connectable_observable.dart';
 export 'package:rxdart/src/observables/observable.dart';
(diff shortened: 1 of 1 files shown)


brianegan:
I'm a bit ambivalent on this issue. I can definitely see how this is useful, but I also wouldn't want that line to grow out of control (exporting StreamController and many other classes, for example).

@frankpepermans Do you have a stronger opinion one way or the other?

frankpepermans:
I think it's best to keep the library export scope to a minimum.

- cleaner for people who are inspecting the source,  no clutter
- there's still a lot of confusion for new users on what is Dart SDK, and what is offered as rxdart.
If you make users import StreamController separately,  then they are perhaps better aware that this class is not rxdart, but rather sdk.

But, we could maybe offer a secondary lib file, i.e. rxdart/rxdart_common where we do include others?

lukepighetti:
I think your trepidation is reasonable, I shall close this PR. Thanks!

PR #337, opened 2019-09-26
add ValueObservable.combineLatestN + tests

lukepighetti:
Quick PR just to see what your appetite is for ValueObservable specific methods to preserve ValueObservable type

### DIFF
--- a/lib/src/observables/value_observable.dart
+++ b/lib/src/observables/value_observable.dart
@@ -1,4 +1,7 @@
+import 'dart:async';
+
 import 'package:rxdart/src/observables/observable.dart';
+import 'package:rxdart/streams.dart';
 
 /// An [Observable] that provides synchronous access to the last emitted item
 abstract class ValueObservable<T> implements Observable<T> {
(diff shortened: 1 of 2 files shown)


brianegan:
I actually really like the idea behind this change. It's something I'd considered in the past as well!

One question about the implementation that I'd love some feedback on: It currently uses `shareValueSeeded` to achieve the effect. That will mean the Observable will become unusable after the final listener unsubscribes to it. Would that be the expected behavior for peeps using this code?

lukepighetti:
I was not aware that was a side-effect of using `shareValueSeeded`. The purpose for using it was to try to preserve `value` at all costs.

brianegan:
Heya @lukepighetti  -- how are ya feeling about this change?

lukepighetti:
If you can provide a quick and clear request for the change I'd be happy to implement it and push

brianegan:
One thing I was considering: I don't think most people want `shareValueSeeded`, which will shut down when there are no more listeners. Rather, I think most folks want an operator that works like `asBroadcastStream` but contains the latest value. I was thinking of adding two new operators:

`asBroadcastValueStream` & `asBroadcastValueStreamSeeded`. These would work exactly like `asBroadcastStream` -- in fact, they'd use that method under the hood, but would also provide access to the latest emitted value.

Perhaps we could use that base for this change?

Issue #345, opened 2019-10-13
switchMap throws exception when original stream is being closed


audkar:
Test to reproduce issue:
```dart
  test('rx.Observable.switchMap stream close after switch', () async {
    final controller = StreamController<int>();

    Observable(controller.stream)
        .switchMap((it) => Stream.fromIterable([it, it]))
        .toList();

    controller.add(1);
    await Future<void>.delayed(Duration(microseconds: 1));
    controller.add(2);

    await controller.close();
  });
```

frankpepermans:
Thanks for the report!

Should be fixed with this PR: https://github.com/ReactiveX/rxdart/pull/347

PR #347, opened 2019-10-14
switchMap: controller closes prematurely

frankpepermans:
As observed here: https://github.com/ReactiveX/rxdart/issues/345

We keep a `rightClosed` flag, which switches to `true` once the secondary `Stream` closes.
However when the primary `Stream` fires a new `event`, causing a new `listen` on a secondary `Stream`, this flag is not being reset to `false`.


frankpepermans:
oh, yes, I've only committed the fix :/

added now

PR #349, opened 2019-10-16
Extension Methods phase 1

brianegan:
### Summary 

  - Introduce `*Extension` extension classes for each `Stream` / `StreamTransformer` that provides a transform method.
  - Move all Rx operators from `Observable` to `ObservableExtensions`
  - Remove all "wrapper" methods from `Observable`
    - Default to `Stream` Implementations
  - Keep constructors on `Observable` class (for now?)
  - Remove Exported Futures & Tests
    - `AsObservable` Future no longer provides utility, since all streams will have Rx methods
    - Min and Max are easily inlined

### Breaking Changes

  - All methods that previously returned an `Observable<T>` now return a plain `Stream<T>`
  - All `Future` implementations removed.

### Discussion

I'd love to hear feedback on this change -- good or bad! This PR will result in some breaking changes, but I've tried to make them as painless as possible for this first phase. If these changes are too drastic, please let me know and we can discuss!

### Rationale

In practice, this PR should make it much easier to use Rx transformations, since they will be available to all streams by importing rxdart. 

```

// Uses the `max` extension method
// Prints 3
Stream.fromIterable([1, 2, 3]).max().then(print); 

// Uses the `onErrorReturn` extenion method
// Prints "Recovery Value"
Stream<String>.error(Exception()).onErrorReturn("Recovery Value").listen(print); 
```

### Background

RxDart is built upon Stream primitives. All operators are implemented as `StreamTransformer` classes, such as `ScanStreamTransformer`. All constructors are implement as regular `Stream` classes, such as `RangeStream`.

The Observable class itself is simply a wrapper around a Stream that makes use of these primitives to add additional methods to the Stream it wraps. This includes the `scan` method, or the `Observable.range` constructor.

With the advent of Dart 2.6, we can take advantage of Static Extension Methods to add the rx transformation methods onto the Stream class, without requiring end-users to wrap their Streams in an Observable!

This means we can remove almost all methods that proxy up to the Stream class, and move the Rx operators to an `extension`.


lib/src/extensions.dart
```diff
@@ -0,0 +1,1456 @@
+import 'dart:async';
+
+import 'package:rxdart/streams.dart';
+import 'package:rxdart/transformers.dart';
+
+import 'observables/connectable_observable.dart';
+import 'observables/replay_observable.dart';
+import 'observables/value_observable.dart';
+
+extension ObservableExtensions<T> on Stream<T> {
```
brianegan: After looking at this extension, I'm actually thinking it might be better to move each of these extensions next to the `StreamTransformer` classes themselves.

brianegan: All extensions that can be moved out have been. Only remaining extensions on this class are sort of "Stragglers" that have no home.

frankpepermans:
Wow great job!

I don't think we should be *too* worried about breaking, this is, to me, a path that we should definitely follow.

Rxdart is an *extension* of Streams, both the source and the api now literately reflect that, making it much easier to understand :)

brianegan:
If we're not *too* worried about breaking changes, what would you and others think about removing the Observable class completely?

At this point, it only provides constructors that could be replaced by the default Stream counterparts. In order to make "phase 1" a bit less harsh when it comes to breaking changes, I was thinking about adding a `@Deprected` warning to all remaining constructors pointing peeps to the base Stream.

E.g. asking people to use `RangeStream` directly in place of `Observable.range`.

frankpepermans:
E.g. asking people to use `RangeStream` directly in place of `Observable.range`.

Ah, so I assume extension methods for factory constructors is not possible? 
If so, then indeed `Deprecating` is a bit more complex, but I agree to point users to the equivalent `Stream` classes then.

I'll load up this branch with 2.6-dev when I find the time to play around :)

brianegan:
> Ah, so I assume extension methods for factory constructors is not possible?

Exactly. We can add `static` methods to an `extension`, but we cannot add `factory` or constructors via an extension.

Even when you add a `static` method to an `extension`, it is not (yet?) visible on the base class. So if you add a static implementation of `just`, you would still have to use `Observable.just(1)` rather than `Stream.just(1)`.

Therefore, it seems to make more sense to point peeps to the base Streams. In fact, it's often less typing to use those base classes as well: `RangeStream` vs `Observable.range`, `NeverStream` is shorter than `Observable.never`, etc.

lib/src/extensions.dart
```diff
@@ -0,0 +1,66 @@
...
+  ///
+  ///     print(min); // prints 1
+  ///
+  /// ### Example with custom [Comparator]
+  ///
+  ///     final stream = Stream.fromIterable(["short", "looooooong"]);
+  ///     final min = await stream.min((a, b) => a.length - b.length);
+  ///
+  ///     print(min); // prints "short"
+  Future<T> min([Comparator<T> comparator]) =>
```
passsy: Add

passsy: Also add `IntObservableExtensions on Stream<int>` for `Future<int> min()`. Same for `double`

brianegan: Are those specific extensions needed? 

`Stream.fromIterable([1, 2, 3]).max().then(print);` works just fine as is (prints 3)?

passsy: The return type right now is `dynamic`. It would be better to return `int` or `double`. Especially because `implicit-casts` will be removed when nnbd lands https://github.com/dart-lang/sdk/issues/31410#issuecomment-510683629

frankpepermans:
Regarding Streams (JustStream, NeverStream, ...)

Thoughts on providing methods instead?

i.e.

import 'package:rxdart/rxdart_streams' as rxs;

var stream = rx.just(1);
var handler = <T>(T event) => rx.just(event);

stream

lib/src/extensions.dart
brianegan: Hrm, I might be missing something here, but the return type of this method is `Future<T>`, not `dynamic`.

Tested with this:

```
test('Should be the correct type', () {
    expect(
      Stream.fromIterable([1, 2, 3]).max(),
      TypeMatcher<Future<int>>(),
    );
  });
```

passsy: It works for types implementing Comparable. But the api doesn't restrict calling `max()` on `Stream<Comparable<T>>`. It results in a runtime crash.

```dart
final allObservables = [Observable.just("a"), Observable.just("b")];
final Observable<String> bestObservable = await Stream.fromIterable(allObservables).max();
// _TypeError (type 'Observable<String>' is not a subtype of type 'Comparable<dynamic>')
```

This can be prevented by declaring the extensions `on Stream<Comparator<T>>`

```dart
extension ComparableExtensions<T extends Comparator<T>> on Stream<T> {
  Future<T> max() =>
      toList().then((List<T> values) => (values..sort()).last);
```

This makes the code snipped above invalid

Screen-Shot-2019-10-17-15-35-56 20

brianegan: Ah, yes -- that is true and always has been for this method -- we'd actually even considered removing them at one point because of this deficiency.

If we want to keep them, I think it's a good idea to restrict it to `Stream<Comparable<T>>`, which would make sure it's type safe and works correctly by default.

Really appreciate your time reviewing 🎉

robrbecker:
I love the direction. Would it be possible to provide a migration tool to update people's code? Take a look at https://github.com/Workiva/dart_codemod or https://comby.dev/

escamoteur:
Why not replace the constructors with either one factory class with static methods?

I guess static extension methods are not possible, are they? 

In general I agree with this direction because advantages overweight the point that the naming is no longer according to rx standards.

brianegan:
@robrbecker I think that's a pretty cool idea, actually! 

For this changeset, I actually don't envision people will have to make *too* many breaking changes. But if we deprecate the constructors that would be the perfect place for such a tool.

brianegan:
@frankpepermans @escamoteur When it comes to constructors on Observable, I personally think we should aim to get rid of them and point folks toward the proper constructor for the class they're interested in.

If you instantiate a `ZipStream`, you know exactly what you're getting. If you used something like `Observable.zip` or `rx.zip`, you actually have to look at the implementation to see that all it's doing is calling the `ZipStream` constructor for you.

In that regard, it feels like an abstraction or indirection without much benefit, other than maybe saving a few characters of typing in some cases. In many (most?) cases, it's actually shorter to simply use the Stream constructor directly and doesn't add any indirection / mental overhead to understand what is happening.

frankpepermans:
Generally, I totally agree, but it's rx, a well known cross platform library.

I'm sure most will know what rx.zip or rx.combineLatest2 returns.

It just feels more Darty too :)

Why don't we do both? Say we provide SomethingStream in the standard package, but also provide a special library where they are presented as above?

I.e. package:rx/rx_methods.dart for example, which needs to be specifically imported.

We could include a builder which just generates this from the sources, including docs?

shinayser:
> Regarding Streams (`JustStream`, `NeverStream`, ...)
> 
> Thoughts on providing methods instead?
> 
> i.e.
> 
> `import 'package:rxdart/rxdart_streams' as rxs;`
> 
> ```
> var stream = rx.just(1);
> var handler = <T>(T event) => rx.just(event);
> 
> stream
>     .switchMap(handler); ...
> ```

No please, **don't do that**! Don't force us to import **rx.dart** everytime with a suffix. This would kill the auto import of dart language (_wich is already a semi-garbage_). 

I give +1 to include those methods inside a class instead of forcing us to use affixes on imports!

frankpepermans:
The suffix is just an example, importing without would simply allow writing

combineLatest2 vs CombineLatest2Stream for example.

I assume auto imports will improve over time, they are indeed painful at the moment.

Personally I like xStream too, but I do see a lot of popular Dart libs exposing methods instead of classes as well.

passsy:
Side note: [ktdart](https://github.com/passsy/kt.dart/tree/master/lib/src/collection) offers top-level functions like `mutableListOf(a, b, c)` as well as constructors `KtMutableList.of(a, b, c)`. I personally prefer the top-level functions but most users I've seen use the factory constructors.

escamoteur:
I would prefer a Factory class may call ist just rx that has factory functions to create the different streams if we can't create static factory extension methods on the Stream type.
This would enable easy discovery through intellisence

rullyalves:
I would find it interesting if static methods and factory builders were moved to functions only, such as:
Observable.just (1)
would turn into:
just (1)

hoc081098:
Vote for top level function like `rxjs`: `of`, `zip`, `timer`, tc

escamoteur:
@brianegan how is the timeline of this change? I have a talk on RxDart on the 2nd December so I would like to know if it makes sense to still explain the Observables

brianegan:
Thanks so much for your feedback everyone! 

@escamoteur I'd like to merge this in the next week or so, or at least deploy a `-dev` version to pub with this work so folks can try it out and report issues.

It seems like the big sticking point comes down to the constructors, but I don't know if we need to make any changes to that part at this very moment to get the benefit of static extensions on all Stream classes.

escamoteur:
@brianegan What do you think to have one StreamBuilder factory class with static methods for Stream creation?

I will then update my presentation to the new API.

escamoteur:
@brianegan What is the current status of this?

brianegan:
@escamoteur I was hoping Dart 2.6 would land stable (past it's due date) and we could publish it as the next version. Since it looks like Dart 2.6 might take a bit longer, I'll go ahead and publish a `-dev` version to pub.

Since folks seem to like the current `Observable` class for easy access / discovery of the available factories, and since it will result in the least number of breaking changes, I'll keep those in place for now. 

We can discuss alternative factory functions / classes in a follow-up "Phase 2" PR when the time comes.

jacobaraujo7:
Static methods instead of Observable factory does not seem to be the best way. Please keep the class Observable factories

brianegan:
Thanks so much for the discussion, all!

  - I've kept `min` and `max` as is, since I didn't want to introduce a functional change there just yet
  - Removed `concatMap` since it's just an alias for `asyncExpand`
  - Removed `ofType`, deprecated in favor of `whereType`
  - Observable class still remains intact! All factories remain in place.

Since this is a a big change, this will not go to the master branch just yet. I will merge this into the `extension-methods` branch. I will use that as a working branch for the next version that will include extension methods, and publish `0.24.0-dev` versions from that branch until the feature looks good.

PR #359, opened 2019-10-24
Fix DeferStream: make streamFactory is called when DeferStream is listened to

### LINKED ISSUE #360: Bug: streamFactory param is called immediately in constructor
Test failed:
```dart
  test('rx.Observable.defer.streamFactory.called', () async {
    var count = 0;

    streamFactory() {
      ++count;
      return Observable.just(1);
    }

    var deferStream = DeferStream(
      streamFactory,
      reusable: false,
    );

    expect(count, 0); // failed here

    deferStream.listen(
      expectAsync1((_) {
        expect(count, 1);
      }),
    );
  });
```

```
package:test_api                   expect
test/streams/defer_test.dart 43:5  main.<fn>

Expected: <0>
  Actual: <1>

Process finished with exit code 1
```

hoc081098:
Fix #360 
Similar #358 
- Bug: `streamFactory` param is called immediately in constructor
- Expected behavior: `streamFactory` should be called when stream is listened to

### DIFF
--- a/lib/src/streams/defer.dart
+++ b/lib/src/streams/defer.dart
@@ -23,11 +23,10 @@ class DeferStream<T> extends Stream<T> {
   DeferStream(Stream<T> streamFactory(), {bool reusable = false})
       : _isReusable = reusable,
         _factory = reusable
-            ? (() => streamFactory())
+            ? streamFactory
             : (() {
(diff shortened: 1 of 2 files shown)


brianegan:
Thanks again!

PR #364, opened 2019-11-05
Fixes BehaviorSubject starting with Error, creates startWithError operator

frankpepermans:
BehaviorSubject erroneously did 'addError' to the controller when called as deferred.

This is fixed by using startWithError instead.

This op is just like startWith, but oviously for Errors instead. Because it might be handy (testing for example), I've decided to add it as a standard op as well.


README.md
```diff
@@ -149,6 +149,7 @@ var myObservable = Observable.combineLatest3(
 - [scan](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/scan.html) / [ScanStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx/ScanStreamTransformer-class.html)
 - [skipUntil](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/skipUntil.html) / [SkipUntilStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx/SkipUntilStreamTransformer-class.html)
 - [startWith](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/startWith.html) / [StartWithStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx/StartWithStreamTransformer-class.html)
+- [startWithError](https://www.dartdocs.org/documentation/rxdart/latest/rx/Observable/startWithError.html) / [StartWithErrorStreamTransformer](https://www.dartdocs.org/documentation/rxdart/latest/rx/StartWithErrorStreamTransformer-class.html)
```
brianegan: Overall, I don't think this should be a public method / transformer.

frankpepermans:
It was actually private first, but upon creating it, I felt it might be useful for some use cases :)

I'd then suggest just removing Observable.startWithError, but then keeping the class, but not exposing it via transformers.dart

So in short, class is public and lives in its own file, but not exposed.
This way you could still use it, if you create a specific path import for it.

README.md
frankpepermans: Removed it (in README and as an Observable operator)

frankpepermans:
@brianegan 
* removed `startWithError` from `Observable`
* removed from README

brianegan:
Sounds good! LGTM :)

PR #368, opened 2019-11-19
Extension methods

brianegan:
Base branch for all extensions method work. This branch will be published as incrementing `-dev.X` releases until it's stable and ready to merge into master. All PRs to improve the extension methods work should be based against this branch until it's merged to master. 

PRs merged thus far:

  - #349 
  - #372 

## Goal

Use extension methods to provide RxDart operators to all Stream classes.

## Effects

Extension methods allow us to make several api improvements:

  1. No longer need to wrap a Stream in an `Observable` class to use Rx operators
  2. No longer need to use the exported `StreamTransformers` -- instead, prefer using the extension methods provided by RxDart
  3. No need for special Observable constructors that internally wrap a Stream in an Observable.
  4. Because of this, we can remove the `Observable` class entirely. Just use `Stream` classes since all operators are available! 
  5. Align more closely with the Streams API. For example, the `Stream.value` constructor now exists, allowing us to deprecate `Observable.just`.

## Before and After examples

|Before|After|
|------|-----|
|`Observable(myStream).flatMap(...)` | `myStream.flatMap(...)`|
|`myStream.transform(FlatMapStreamTransformer(...))`| `myStream.flatMap(...)`|
|`Observable.fromIterable([1, 2, 3])`|`Stream.fromIterable([1, 2, 3])`|
|`Observable.concat(...)`|`ConcatStream(...)`|


lib/src/transformers/timestamp.dart
```diff
@@ -78,3 +78,17 @@ class Timestamped<T> {
     return 'TimeStamp{timestamp: $timestamp, value: $value}';
   }
 }
+
+/// Extends the Stream class with the ability to
```
frankpepermans: Incomplete docs

lib/src/transformers/time_interval.dart
```diff
@@ -97,3 +97,19 @@ class TimeInterval<T> {
     return 'TimeInterval{interval: $interval, value: $value}';
   }
 }
+
+/// Extends the Stream class with the ability to wrap each item emitted by the
+/// source Stream in a [Timestamped] object that includes the emitted item
```
frankpepermans: [Timestamp] is from timestamp.dart, here we use [TimeInterval]

lib/src/transformers/ignore_elements.dart
```diff
@@ -42,3 +42,18 @@ class IgnoreElementsStreamTransformer<T> extends StreamTransformerBase<T, T> {
     });
   }
 }
+
+/// Extends the Stream class with the ability to skip, or ignore, data events.
+extension IgnoreElementsExtension<T> on Stream<T> {
+  /// Creates a Stream where all emitted items are ignored, only the error /
```
frankpepermans: I wonder if we still need ignoreElements, given that Stream.drain exists?

lib/src/transformers/flat_map.dart
```diff
@@ -98,3 +98,41 @@ class FlatMapStreamTransformer<T, S> extends StreamTransformerBase<T, S> {
...
+  ///
+  /// Use case: you may have an API that returns a list of items, such as
+  /// a Stream<List<String>>. However, you might want to operate on the individual items
+  /// rather than the list itself. This is the job of `flatMapIterable`.
+  ///
+  /// ### Example
+  ///
+  ///     RangeStream(1, 4)
+  ///       .flatMapIterable((i) =>
+  ///         Stream.fromIterable([[]i])
```
frankpepermans: I suppose this should be Stream.fromIterable([i]) instead?

lib/src/transformers/time_interval.dart
frankpepermans: ...this is actually a legacy doc error, nothing to do with this PR

lib/src/transformers/ignore_elements.dart
brianegan: It's actually a fair point. Since this PR doesn't really introduce any functionality changes (mostly just rename changes which are easy to handle with a codemod), what would you think of deprecating this method as part of this PR to give folks time to adjust their code?

lib/src/transformers/time_interval.dart
brianegan: Good catch, I had mixed up the `TimeStamped` and `TimeInterval` docs!

lib/src/transformers/ignore_elements.dart
brianegan: Marked `ignoreElements` as deprecated, then we can fully remove it in an upcoming release!

frankpepermans: Sure, should we do the same with just?

brianegan: Since `Stream.value` works the exact same way as `Observable.just` and can easily be found/replaced or updated with a codemod, and because keeping `Observable.just` would require us to keep the `Observable` class in general, I think it makes sense to remove it as part of this PR.

hoc081098:
Please export `connectable_stream.dart`, `value_stream.dart`

brianegan:
Very good catch @hoc081098 -- thanks!

PR #372, opened 2019-11-23
Extension methods phase 2

brianegan:
Now that we have all the extension methods in place, it's time to talk about a larger breaking change: Removing the `Observable` class.

I understand there's some hesitation around this change, but I'd ask folks to read the following information and take a look at the code changes -- personally, I think these changes will make your code and RxDart code much leaner and less confusing in the long run (as can be seen by the number of lines removed vs added!), but I'd love to hear feedback as to why you might disagree!

## Rationale

Many folks might not know that the `Observable` class is just a wrapper class.  It does not exhibit any special behavior compared to Streams, since Observable *is* a Stream! With extension methods, it's not actually needed at all :)

However, many folks believe the `Observable` class we provide will work exactly like other Rx `Observables`, which terminate on error and can be subscribed to more than once. Given this is a common source of confusion, I propose we remove the `Observable` class altogether.

## Background

In reality, the `Observable` class serves only two purposes:

  - To wrap the source Stream provided to the constructor. In the past, this allowed us to add additional methods to the wrapped Stream.
  - To provide factory constructors and static factories. These were used to create Streams provided by the standard library or RxDart, ensuring they were wrapped in an Observable.

With extension methods, users no longer need to wrap Streams in an Observable to use the additional methods provided by RxDart!

Therefore, at this point, the only purpose of the Observable class is to provide Stream factories. However, even these these Stream factories no longer need to wrap the Streams they create in an `Observable` class! For example, you no longer need to use `Observable.fromIterable`, you could just use `Stream.fromIterable` and the functionality will be identical.

## Solution

Therefore, I suggest we remove the `Observable` class entirely and move all factories to a new `Rx` class. Remove all constructors that simply create Streams from the Standard library, e.g. remove `Observable.fromIterable` and replace it with `Stream.fromIterable`

From listening to feedback from all you fine folks (thanks to all who wrote us!), it seems people want the factories to stick around. My proposal: Move all factories to a new class called `Rx` and make them all static. This will mean creating streams provided by this library will still be nice and concise, yet easily discoverable.

### Before and After Comparisons

|Before|After|
|------|-----|
|`Observable.just(1)`|`Stream.value(1)`|
|`Observable.error(Exception())`|`Stream.error(Exception())`|
|`Observable.fromIterable([1, 2, 3])`|`Stream.fromIterable([1,2,3])`|
|`Observable.combineLatest2`|`Rx.combineLatest2`|
|`Observable<int>.concat`|`Rx.concat<int>`|

## Migration Strategy

Since this would require a lot of changes, I plan on writing a [codemod](https://pub.dev/packages/codemod) (thanks to robrbecker for the suggestion!). The codemod would go through your codebase and update all necessary constructors.

I also plan on doing a search for blog posts and articles on RxDart and plan on contacting authors asking them if they could update the code snippets to the latest version.

## List of Changes

  * Remove the Observable class. With extensions, you no longer need to wrap Streams in an Observable!
  * Convert all factories to static constructors to aid in discoverability of Stream classes
  * Move all factories to an `Rx` class.
  * Remove `Observable.just`, use `Stream.value`
  * Remove `Observable.error`, use `Stream.error`
  * Remove all tests that check base Stream methods
  * Subjects and *Observable classes extend Stream instead of base Observable
  * Rename *Observable to *Stream to reflect the fact they're just Streams.
    * `ValueObservable` -> `ValueStream`
    * `ReplayObservable` -> `ReplayStream`
    * `ConnectableObservable` -> `ConnectableStream`
    * `ValueConnectableObservable` -> `ValueConnectableStream`
    * `ReplayConnectableObservable` -> `ReplayConnectableStream`


raveesh-me:
Here is my take on this:

1. From a purely flutter point of view it makes perfect sense. 

2. From a full-stack developer's point of view who switches between multiple languages using Rx as the common paradigm..  I would like us to keep calling streams as observable in dart.

brianegan:
Thanks Raveesh! Could you please describe in a bit more detail why calling them observables is helpful? Is it just that seeing the name, you generally know what they do, whereas if they were called Streams that wouldn't be the case?

brianegan:
Feedback from twitter via Ana Betts and in-line response / follow-up questions:

  > RxJS mostly does this, it only exists to do Observable.create. 

  > I don't understand why you'd rename this class to Rx - it seems like extra breaking changes and confusion for no benefit. You really need to consider the ramifications of destroying the RxDart ecosystem like this

The idea behind renaming to `Rx` came from two thoughts: How can we make the static factories as short as possible, and second, how can we help people understand these aren't traditional Rx observables, which has been a big source of confusion, but rather just Dart Streams. 

I think the big question behind that particular change: Is it worth making the change to help current and future users of RxDart understand these aren't normal Observables? 

  > Everyone googling RxDart will get information that's similar but they'll try it and it doesn't work, people learning RxDart now need to learn both streams as a Concept (already hard!) *And* translate any tutorial / blogs to http://RxDart.New

Could you describe a bit more why you think it'd destroy the community? My personal impression of big changes, from far larger ecosystems like RxJS, was that the community adopted the new changes quite quickly.

I was also hoping if we provided a migration tool it would get people's code-bases up to speed, but perhaps need to rethink the ramifications of the wider set of tutorials.

  > Do the extension method improvements but:
  >   1. Make it as consistent as possible with what you had before
  >   2. Make it as consistent as possible with other Rx implementations

Agree 100% -- that's exactly why I split this work out of "phase 1", which implemented extension methods with as few changes to the Observable class as possible. 

This proposed "Phase 2" is certainly more radical, and if folks think it's too big a change we do not need to merge any of this work :)

In general, we certainly aim to be as close to other implementations as possible, but that's often quite tricky when you start comparing different implementations and seeing how different they are from one another. 

Therefore, our goal has generally been to stay true to the ReactiveX spec while making the library as idiomatic to Dart as possible, which is why we don't even implement our own `Observable` and simply extend `Stream`.

simolus3:
I welcome this change, but I'm then wondering how rxdart is different to an alternative package from the Dart team.

The [stream_transform](https://pub.dev/packages/stream_transform) package from the Dart team provides many transformations as extension methods on `Stream`. Of course, rxdart provides more features and is arguably the more popular package. In the past, rxdart has had a clear distinction from stream_transform and was more familiar for people coming from other rx libraries. Now that rxdart is getting rid of the observable wrapper and stream_transform started to use extension methods as well, we have two popular packages doing a very similar thing in a very similar way. 

RxDart still provides operators not available in stream_transform and of course still contains the behavior/replay subjects. But to avoid duplication, maybe rxdart could depend on and export apis from the stream_transform package instead of re-implementing them? This could help reduce code size and make it more obvious which package to chose for users: stream_transform for simpler and more lightweight transformations, and rxdart for a batteries-included solution providing more operators and subjects.

brianegan:
@simolus3 Thanks for the feedback.

I completely agree that the two libraries are quite similar, and to be honest they always have been -- you could always import only the `transformers.dart` file from RxDart to use only the StreamTransformers provided by this library.

I don't think it's a bad idea to have closer collaboration with that library, but many of their transformers are functionally different than what RxDart provides, so that would actually be a much more difficult change for end-users to make.

talamaska:
My opinion might be not a popular one. I'm coming in flutter from Angular (not dart), I thought I was understanding rxjs and a bunch of its operators, but I was really confused with rxdart. There are some critical differences between the two, you mentioned some those. But the core of the problem is that dart already have streams with their own logic, so just wrapping it and naming things is not making it the same, just more confusing. No Hot/cold observables, streams need to broadcast in order to be listenable from multiple points - all that stuff. I like the differentiation between the Subjects, that rxdart gives me, avoiding the stream controller usage. For me its hard to understand. Its not just rxdart, dart unlike angular, has some things Futures/Promises, streams, async is used in both? Async, async*, yield, yield* I'm confused, isn't that like generators in ES7. In the end I just want ngrx for flutter :). I'm supporting your decision removing the Observable, since it's not the same Observable as in rxjs.

CONTRIBUTING.md
```diff
@@ -14,19 +14,12 @@ If you'd like to add a feature or fix a bug, we're more than happy to accept pul
...
-  - If you would like to make a bigger / fundamental change to the codebase, please file a lightweight example PR / issue, or contact us in [Gitter](https://gitter.im/ReactiveX/rxdart) so we can discuss the issue.
-
-## Advice when adding a new factory
-
-  - Extend from `Stream` so it can be constructed outside the scope of the Observable class
-  - Add the new `Stream` to the exported `rx_streams` library
-  - Ensure the stream properly enforces the single-subscription contract
-  - Ensure the stream closses properly
-  - Add new tests to `tests/rxdart_test.dart`
+  - If you would like to make a bigger / fundamental change t
```
felangel: Minor but looks like the docs got cut off

example/flutter/github_search/lib/search_result_widget.dart
```diff
@@ -4,8 +4,7 @@ import 'package:github_search/github_api.dart';
 class SearchResultWidget extends StatelessWidget {
   final List<SearchResultItem> items;
 
-  SearchResultWidget({Key key, @required this.items})
-      : super(key: key);
+  SearchResultWidget({Key key, @required this.items}) : super(key: key);
```
felangel:
```suggestion
  const SearchResultWidget({Key key, @required this.items}) : super(key: key);
```

BaranMichal25:
> However, many folks believe the Observable class we provide will work exactly like other Rx Observables, which terminate on error and can be subscribed to more than once. (...) I propose we remove the Observable class altogether.

Instead of removing Observable, can't we go in direction of actually having Observable class working exactly as other Rx Observables?

I'm afraid that currently this project is going in direction of "useful extensions and operators for Streams" instead of being truly ReactiveX library with APIs aligned with official [documentation](http://reactivex.io/documentation/observable.html).

hoc081098:
@BaranMichal25 🤣🤣🤣

frankpepermans:
Since Dart streams are so tightly integrated into pretty much everything, I honestly believe it's not a good idea to create `Observable` from the ground up.

Users would need to learn both core Dart `Streams` and Rx `Observables`, because since Dart Streams are tightly integrated into the SDK, there's no way of avoiding them inside an application.

I'd hate to have a codebase where I would see `Streams` and `Observables` used throughout, having to know how they both work, too many gotcha's to keep in mind.

This library has always been "useful extensions and operators for Streams" really, and the `Observable` name has thrown people off so many times, even though we state that Observable is if fact nothing more than an extended Dart Stream.

We've had many open issues about that difference to Rx, and we're not happy with it either!
But, if you come to accept the differences and learn Dart Streams beforehand, then rxdart is a very convenient library where you do find the same operators and concepts found in any Rx implementation.

This PR finally allows us to drop Observable, 
I don't feel that it should drop Rx as well, once the Dart Stream hurdle is taken, I do believe that this library feels very familiar as you progress into Dart.

anaisbetts:
> Could you describe a bit more why you think it'd destroy the community? My personal and completely ill-informed impression of big changes, from far larger ecosystems like RxJS, was that the community adopted the new changes quite quickly.

I would not characterize this transition as smooth at all, as someone who worked with the RxJS maintainers during the RxJS 4 => 5 => 6 years, this was not at all easy for many people. A *lot* of people were confused and frustrated, and while a lot of these breaking changes were done because they were necessary (i.e. 5 => 6 changed a ton of syntax to make tree-shaking work), it was super difficult for the community, even with RxJS doing its best to help by making compatibility shims.

As experts, we forget that Rx is Hard. It's fundamentally difficult for new people to grasp, and if they have to deal with blog posts that don't work, and translating method names _on top_ of trying to figure out how to conceptually understand streams, it's really going to make learning RxDart difficult.

Even though the change here is totally worth it because extension methods are a Game Changer, we have to be careful about "Mental model breaking changes" - right now, people know that to create an Observable, you start with the Observable class. Even though internally, Observables are just Streams, from a Conceptual Perspective, people are going to go looking for Observable, because that's what Rx Is™

frankpepermans:
@anaisbetts I agree, this is a tough discussion and we've been having it for quite a while.

Isn't it though more frustrating to load up rxdart, see that it just has Observable, then start writing code only to find out at some point that Observable isn't really Observable?

- People who are unfamiliar with the concept of Streams will need to start learning it anyway.
- People already familiar with Dart Streams, but not Rx, will wonder why Observable is there, because it just looks like a regular Stream.
- People coming from rxjs or rxjava to Dart and rxdart will encounter Dart Streams, start looking for Observable instead, and then initially assume that they can just forget about learning Dart Streams for the moment.

RollyPeres:
TBH removing Observable is the best thing you could do. Dart Streams are not Observables. I've worked quite a bit with RxJS and Rx.Net but when using RxDart I just feel that I'm in a different universe which should not contain Observables but Streams only. The only things they both share are the reactive nature and some similar operators, but they really do feel like two step brothers sharing the same mom. This package is more like `StreamExtensions` rather than `RxDart`.

raveesh-me:
> Thanks Raveesh! Could you please describe in a bit more detail why calling them observables is helpful? Is it just that seeing the name, you generally know what they do, whereas if they were called Streams that wouldn't be the case?

Just the familiarity bits. But if by the nature of Streams and the way Flutter is built to have streams everywhere, with basically minimum support for observable from `StreamBuilder` and `StreamProvider`, it makes much more sense to let go off of observable.

As a team through the few apps that we have built, we rely mostly and a lot on `BehaviorSubject` and have never once had the necessity to use `Observable`.

I just home ES gets streams soon 👅

brianegan:
Thanks again to all for the feedback! 

@talamaska 

  > But the core of the problem is that dart already have streams with their own logic, so just wrapping it and naming things is not making it the same, just more confusing. 

<3 We hear ya, feel the same way, and that's the big motivator for me behind this change.

@anaisbetts 

  > As experts, we forget that Rx is Hard. It's fundamentally difficult for new people to grasp, and if they have to deal with blog posts that don't work, and translating method names on top of trying to figure out how to conceptually understand streams, it's really going to make learning RxDart difficult.

Completely agree with this. Streams are hard, and we can't forget this. I also have a lot of sympathy for outdated RxJS tutorials as I've certainly run into those myself.

I don't think we'll be able to fix everything, of course, but since I think the Dart community is still relatively small, what do you think of the idea of us doing a search for rxdart and contacting blog post authors to ask them to update their posts to the latest syntax?

@RollyPeres Haha, we've even talked about renaming this project in the past and it's something I've definitely thought about again from time to time: https://github.com/ReactiveX/rxdart/issues/95

That's a really tough one, though.

@raveesh-me Thanks for the extra info!

CONTRIBUTING.md
brianegan: Oh my, good catch!

example/flutter/github_search/lib/search_result_widget.dart
brianegan: Thanks!

brianegan:
Hey all -- wanted to say thanks again for the feedback! 

Overall, from listening to feedback here and on different platforms, I feel quite confident this is the right direction for the library to take. 

I really appreciate those who disagree or are concerned with such a big change, but there's no getting around the fact that `Observable` is a big source of confusion for users. While the immediate transition might be painful, I really think it's going to be helpful for the community.

I've also been working on a `codemod` which I've used to successfully upgrade a few personal projects all the tests in RxDart to work with the new format. I'll figure out how to publish that package soon!

talamaska:
Looking forward for the codemod

Issue #370, opened 2019-11-22
Add takeWhileInclusive operator


shinayser:
As discussed here:

> `closeAfterEmitting` does not exist, but it might make sense for it to exist since other Rx implementations have something along this line: `takeWhile(myPredicate, inclusive: true)`. Since `takeWhile` comes from the Stream Interface, we'd have to create a new method called `takeWhileInclusive` or something.

_Originally posted by @shinayser in https://github.com/ReactiveX/rxdart/issues/369#issuecomment-557580054_

frankpepermans:
So basically this is a `takeWhile`, but it also adds the next event before closing?

So imagine:
Stream.fromIterable([0, 1, 2, 3, 4, 5, 6, 7])
      .takeWhileInclusive((i) => i < 5)
      .listen(print); // 0, 1, 2, 3, 4, **5** done

Should include also have effect the other way around? Take:
Stream.fromIterable([0, 1, 2, 3, 4, 5, 6, 7])
      .takeWhileInclusive((i) => i > 1)
      .listen(print); // **1**, 2, 3, 4, 5, 6, 7 done

What about this one?
Stream.fromIterable([0, 1, 2, 3, 4, 5, 6, 7])
      .takeWhileInclusive((i) => i > 1 && i < 5)
      .listen(print); // **1**, 2, 3, 4, **5** done

shinayser:
Tricky question... 

There are 2 possible ways of implementing it. 

### Case A. 
Closing immediatly after receiving the element that "falses" the condition.
So, in the second example it should print: `1, close`
In the bottom example it should print `1, done`.

### Case B.
 Ignore elements until it finds an element that matches the condition. Then, after finding it, it starts respecting to the case **A**. 
So in the second example, it should print: `2, 3, 4, 5, 6, 7, done`
In the bottom it should print: `2, 3, 4, 5, done`

Perhaps a boolean parameter to the function could inform what kind of behavior the user wants. Something like that:

`myStream.takeWhileInclusive( (i) => i > 1, waitMatch: true);`

You could add our examples to the method documentation. I think it will be great that way :)

frankpepermans:
For **case B**, something like `ignoreWhile` might be better?

```
Stream.fromIterable([0, 1, 2, 3, 4, 5, 6, 7])
      .ignoreWhile((i) => i < 2)
      .listen(print); // 3, 4, 5, 6, 7 done
```

```
[...]

PR #383, opened 2019-12-19
Add takeWhileInclusive operator

hoc081098:
#370

### DIFF
--- a/lib/src/transformers/take_while_inclusive.dart
+++ b/lib/src/transformers/take_while_inclusive.dart
@@ -1,0 +1,92 @@
+import 'dart:async';
+
+/// Emits values emitted by the source Stream so long as each value
+/// satisfies the given test. When the test is not satisfied by a value, will
+/// emit it and complete.
+///
+/// ### Example
+///
+///     Stream.fromIterable([2, 3, 4, 5, 6, 1, 2, 3])
+///       .transform(TakeWhileInclusiveStreamTransformer((i) => i < 4))
+///       .listen(print); // prints 2, 3, 4
+class TakeWhileInclusiveStreamTransformer<T>
+    extends StreamTransformerBase<T, T> {
+  final StreamTransformer<T, T> _transformer;
+
+  /// Constructs a [StreamTransformer] which forwards data events while [test]
+  /// is successful, and includes last event that caused [test] to return false.
+  TakeWhileInclusiveStreamTransformer(bool Function(T) test)
+      : _transformer = _buildTransformer(test);
+
+  @override
+  Stream<T> bind(Stream<T> stream) => _transformer.bind(stream);
+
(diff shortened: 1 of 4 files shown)


frankpepermans:
I was just going to open a PR for exactly this, but you beat me to it hehe :)

I'll check it asap, many thanks!

lib/src/transformers/take_while_inclusive.dart
```diff
@@ -0,0 +1,92 @@
...
+          bool satisfies;
+
+          subscription = input.listen(
+            (data) {
+              try {
+                satisfies = test(data);
+              } catch (e, s) {
+                controller.addError(e, s);
+                // The test didn't say true. Didn't say false either, but we stop anyway.
+                controller.close();
```
frankpepermans: This line should be removed:
upon listening, one can choose to set cancelOnError: false

hoc081098: This behavior is same as `Stream.takeWhile`, https://dartpad.dartlang.org/8949740dda7ba576ed0241089a5041da

frankpepermans: Heh you're right :) 
In that case it's best to do the same thing indeed!

lib/src/transformers/take_while_inclusive.dart
```diff
@@ -0,0 +1,92 @@
+import 'dart:async';
+
+/// Emits values emitted by the source Stream so long as each value
+/// satisfies the given test. When the test is not satisfied by a value, will
+/// emit it and complete.
```
frankpepermans: Just wording here: I'd use "... it will emit this value as a final event and then complete "

lib/src/transformers/take_while_inclusive.dart
```diff
@@ -0,0 +1,92 @@
...
+    });
+  }
+}
+
+/// Extends the Stream class with the ability to take events while they pass
+/// the condition given and include last event that doesn't pass the condition.
+extension TakeWhileInclusiveExtension<T> on Stream<T> {
+  /// Emits values emitted by the source Stream so long as each value
+  /// satisfies the given test. When the test is not satisfied by a value, will
+  /// emit it and complete.
```
frankpepermans: same as above

lib/src/transformers/take_while_inclusive.dart
hoc081098: Fixed :))

lib/src/transformers/take_while_inclusive.dart
hoc081098: Fixed :))

brianegan:
Thanks so much, @hoc081098, and sorry about the delay! I'll merge this in :)

PR #385, opened 2019-12-21
[Simplify implementation] concatWith, mergeWith, zipWith

hoc081098:
Remove StreamTransformer.fromBind, return Stream directly


frankpepermans:
Looks good, thanks!

Issue #389, opened 2020-01-07
doOnListen converts to broadcast stream and is called on every listen, not just the first.


smkhalsa:
The docs for `doOnListen` state that it 

> Invokes the given callback function when the stream is first listened to.

However, when I run the following...

```
final controller = StreamController();

final stream = controller.stream.doOnListen(() {
  print('listening');
});

stream.listen((data) {});
stream.listen((data) {});
```

The string gets printed twice.

> listening
> listening

This is unexpected for two reasons:
1. The stream has been converted to a broadcast stream, allowing multiple listens
2. The doOnListen callback gets called on each listen, not just on the first as described in the docs.

smkhalsa:
@brianegan is this the expected behavior?

hoc081098:
Haha. `stream` is single subscription stream

frankpepermans:
Hmmm it shouldn't become a broadcast Stream :/

The fact that *every* listen triggers the handler, is as intended though, we need to improve the docs here.

frankpepermans:
While writing some more tests for `doOnListen`, it seems that our implementation here is broken:

The following test hangs on the second listen:

```
test('doOnListen correctly allows subscribing multiple times on a broadcast stream', () {
      final controller = StreamController<int>.broadcast();
      final stream = controller.stream.doOnListen(() {
        // do nothing
      });

      stream.listen(null, onDone: () => print('done A'));
      stream.listen(null, onDone: () => print('done B')); // is being ignored

      expectLater(stream, emitsDone);
      expectLater(stream, emitsDone);

      controller.close();
    });
```
We keep a `subscriptions` `Map` internally to avoid listening to a `Stream` more than once for `doOnListen`, but this causes the second `onDone` method to be ignored.

smkhalsa:
> The fact that every listen triggers the handler, is as intended though, we need to improve the docs here.

Ok, yes, the docs should be updated then.

If doOnListen cannot be used to invoke a callback only on the first listen, is there another operator that allows for this?

frankpepermans:
No, any chance you can just capture that inside your handler, using a flag for example?

frankpepermans:
should be fixed in 0.24.0-dev.1

PR #390, opened 2020-01-08
issue/389

frankpepermans:
@brianegan see https://github.com/ReactiveX/rxdart/issues/389

My fix looks very, very hacky, maybe we should think about a different approach for `doOnListen`?

- rewrites the subscribe time, so that single subscription Streams throw correctly when listening multiple times.
- for broadcast Streams, introduce a *dummy* listener, which allows external listeners to receive events correctly.

For the latter, this is the test that fails with our current master, but works when using the dummy approach:

```
test(
        'doOnListen correctly allows subscribing multiple times on a broadcast stream',
        () {
      final controller = StreamController<int>.broadcast();
      final stream = controller.stream.doOnListen(() {
        // do nothing
      });

      expectLater(stream, emitsDone);
      expectLater(stream, emitsDone);

      controller.close();
    });
```

### DIFF
--- a/lib/src/utils/notification.dart
+++ b/lib/src/utils/notification.dart
@@ -32,7 +32,7 @@ class Notification<T> {
   /// Constructs a [Notification] which, depending on the [kind], wraps either
   /// [value], or [error] and [stackTrace], or neither if it is just a
   /// [Kind.OnData] event.
-  Notification(this.kind, this.value, this.error, this.stackTrace);
+  const Notification(this.kind, this.value, this.error, this.stackTrace);
 
   /// Constructs a [Notification] with [Kind.OnData] and wraps a [value]
   factory Notification.onData(T value) =>
@@ -40,7 +40,7 @@ class Notification<T> {
 
   /// Constructs a [Notification] with [Kind.OnDone]
   factory Notification.onDone() =>
-      Notification<T>(Kind.OnDone, null, null, null);
+      const Notification(Kind.OnDone, null, null, null);
 
   /// Constructs a [Notification] with [Kind.OnError] and wraps an [error] and [stackTrace]
   factory Notification.onError(dynamic error, StackTrace stackTrace) =>
--- a/test/transformers/do_test.dart
+++ b/test/transformers/do_test.dart
@@ -198,13 +198,10 @@ void main() {
       final streamA = Stream.value(1).transform(transformer),
           streamB = Stream.value(1).transform(transformer);
 
-      streamA.listen(null, onDone: expectAsync0(() {
-        expect(callCount, 2);
-      }));
+      await expectLater(streamA, emitsInOrder(<dynamic>[1, emitsDone]));
+      await expectLater(streamB, emitsInOrder(<dynamic>[1, emitsDone]));
 
-      streamB.listen(null, onDone: expectAsync0(() {
-        expect(callCount, 2);
-      }));
+      expect(callCount, 2);
     });
 
     test('throws an error when no arguments are provided', () {
@@ -248,8 +245,8 @@ void main() {
                 ..cancel();
         },
         onError: expectAsync2(
-          (Exception e, [StackTrace s]) => expect(e, isException),
-        ),
+            (Exception e, [StackTrace s]) => expect(e, isException),
+            count: 2),
       );
 
       Stream.value(1)
@@ -290,20 +287,6 @@ void main() {
     });
 
     test(
-        'doOnListen correctly throws when subscribing multiple times on a single subscription stream',
-        () {
-      final controller = StreamController<int>();
-      final stream = controller.stream.doOnListen(() {
-        // do nothing
-      });
-
-      expectLater(stream, emitsDone);
-      expectLater(stream, emitsError(TypeMatcher<StateError>()));
-
-      controller.close();
-    });
-
-    test(
         'doOnListen correctly allows subscribing multiple times on a broadcast stream',
         () {
       final controller = StreamController<int>.broadcast();
@@ -311,10 +294,10 @@ void main() {
         // do nothing
       });
 
-      expectLater(stream, emitsDone);
-      expectLater(stream, emitsDone);
-
       controller.close();
+
+      expectLater(stream, emitsDone);
+      expectLater(stream, emitsDone);
     });
   });
 }
--- a/test/transformers/do_test.dart
+++ b/test/transformers/do_test.dart
@@ -288,5 +288,33 @@ void main() {
             ..pause()
             ..resume();
     });
+
+    test(
+        'doOnListen correctly throws when subscribing multiple times on a single subscription stream',
+        () {
+      final controller = StreamController<int>();
+      final stream = controller.stream.doOnListen(() {
+        // do nothing
+      });
+
+      expectLater(stream, emitsDone);
+      expectLater(stream, emitsError(TypeMatcher<StateError>()));
+
+      controller.close();
+    });
+
+    test(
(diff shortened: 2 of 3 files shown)


frankpepermans:
@brianegan I've decide to rewrite the Transformer in the end, lemme know what you think.

test/transformers/do_test.dart
```diff
@@ -248,8 +245,8 @@ void main() {
                 ..cancel();
         },
         onError: expectAsync2(
-          (Exception e, [StackTrace s]) => expect(e, isException),
-        ),
+            (Exception e, [StackTrace s]) => expect(e, isException),
+            count: 2),
```
frankpepermans: I had to change this test, it now receives 2 Exceptions, and it only received 1 before. Not sure if that is OK.

lib/src/utils/notification.dart
```diff
@@ -32,15 +32,15 @@ class Notification<T> {
...
+  const Notification(this.kind, this.value, this.error, this.stackTrace);
 
   /// Constructs a [Notification] with [Kind.OnData] and wraps a [value]
   factory Notification.onData(T value) =>
       Notification<T>(Kind.OnData, value, null, null);
 
   /// Constructs a [Notification] with [Kind.OnDone]
   factory Notification.onDone() =>
-      Notification<T>(Kind.OnDone, null, null, null);
+      const Notification(Kind.OnDone, null, null, null);
```
frankpepermans: tiny optimization to const here

frankpepermans:
@brianegan :wave:

brianegan:
Oh my, sorry @frankpepermans  -- missed this one! At a conference (about to talk about Rx!) -- I'll check it out shortly :)

brianegan:
Oh, boy -- this one is tough. Sorry it took me a while to look at this one, it's more complex than I initially expected. I think there are two things going on here, and it might help to separate them:

  1. When should onListen actually be called?
  2. The Stream should not be converted from a Single-Subscription to a Broadcast Stream

Right now, we're sort of defining our own behavior for `onListen`, which is different from how it normally works. 

By default, Single-Subscription StreamControllers, `onListen` will be called the first time it's listened to and cannot be called again since it'd be an error to listen to the Stream again.

For Broadcast streams, `onListen` will be called any time the number of listeners goes from `0` to `1`. Overall, I don't find this behavior very intuitive, as I originally expected it to be called only 1 time (when it's first listened to) or N times, where N is the number of listeners.  However, it is the "standard."

Our current implementation does its own thing: We call the `onListen` callback N times, where N is the number of listeners. Overall, I personally think it's the most intuitive, but it breaks with the StreamController standards.

However, in addition to kind of converting a Single-Subscription Stream into a Broadcast Stream, the conversion exposes another issue: We're currently performing the transformation on every listen, which results in the `doOnListen` callback being called N times. BUT -- we're only forwarding the events from the first `onListen` to the first listener, so it's not really a broadcast Stream, but just buggy haha. For example, this only prints "1, 2" instead of "1, 1, 2, 2"

```dart
test('this transformer is pretty busted', () {
    final controller = StreamController<int>();

    final stream = controller.stream.doOnListen(() {
      print('listening');
    });

    stream.listen(print);
    stream.listen(print);

    controller.add(1);
    controller.add(2);
  });
```

Right now, we're using the default `StreamTransformer` constructor to create our StreamTransformer. We did this so that `onListen` can run every time the Stream is listened to, but it also causes the issues we're seeing here.

It's really tricky because the `onListen` behavior is really defined by the StreamController class, and it's just a bit confusing as is. I worry if we add another definition of what "onListen" means, which we've done with our current implementation, it will get even more confusing.

Overall, I wonder if we should try to simplify the implementation, so that `onListen` works exactly as it does for normal StreamControllers, rather than trying to make this fix work? 

To make it work like normal StreamController behavior, we could extend `StreamTransformer` and implement `bind` directly so the original Stream is transformed only once, not on every `listen`. This would prevent us from recreating the StreamController on every listen (meaning single-subscription streams would work as they always have) and the `onListen` functionality would reflect the normal `StreamController` behavior as well.

Sorry -- this is a wall of text, one of those really tricky situations that might be easier to work through by pair programming haha.

frankpepermans:
No worries,

I recall this PR does fix the buggy behavior,
But I need to dive in again, been a while ;)

frankpepermans:
@brianegan Ok, dove back into this PR and took another look:

I've made some changes and do believe that this fix does make sense:
the problem is not with onListen here, I've started out by going back to the default way that we build our `StreamTransformers`, i.e.

- creating a `controller`
- subscribing to the incoming Stream inside the controller's `onListen` handler
- returning a `subscription` to this controller

This way, our issue with single / broadcast is resolved, that is multiple listens will throw the correct `exception` on single ones, while it just works for broadcast ones.

The problem now was handling `onData` correctly, my initial PR did some magic by keeping track of event indices, which I was not too happy about.
The change now does rid of this, and instead simply keeps an internal List of which input Streams are already handling `onData`.

Because each new listen invokes `onListen`, without the above check all `onData` handlers would be called N amount of times (where N is the amount of listeners).

I do wish we had better options regarding Dart Streams, a lot of its inner workings are private implementations and just the way they work is counter intuitive to what other Stream implementations do, but that's a different story ;)

brianegan:
Man, this is just a hard one haha... I see a couple of potential issues with this solution:

  1. Is it really reusable? Every time the StreamTransformer is applied to a new Stream, the `StreamSubscription` will be wiped out and track the new Stream instead.  It's weird, because the reusable tests pass on CI, but do not pass on my machine :/
  2. The `Single Subscription Error` is thrown too Late.

For example, this test passes with a normal Single Subscription Stream:

```dart
test('doOnListen accidental broadcast', () {
      final controller = StreamController<int>();

      controller.stream.listen((data) {});
      expect(() => controller.stream.listen(null), throwsStateError);

      controller.add(1);
    });
```

But it fails when using the transformer, because the error is thrown too late:

```dart
    test('doOnListen accidental broadcast', () {
      final controller = StreamController<int>();

      final stream = controller.stream.doOnListen(() {
        print('listening');
      });

      stream.listen((data) {});
      expect(() => stream.listen(null), throwsStateError);

      controller.add(1);
    });
```

The order of operations:

  1. User listens first time
  2. StreamTransformer (ST) `onListen` called.
  3. ST onListen creates a StreamController (SC)
  4. The SC defines it's own onListen, which FIRST runs the `DoStreamTransformer` `onListen` callback provided by the user. Then the SC onListen callback listens to the source Stream
  5. We listen to the SC and return the Subscription, triggering the source listen.
  5. Everything is good at this point!
  6. A new Listener Emerges.
  7. ST onListen is called again
  8. It creates a new SC
  9. Which calls the `DoStreamTransformer` `onListen` 
  10. Then tries to listen to the Stream
  11. Oops! Now it throws the State error! (Later than expected).

Therefore, it's not directly to do with ST `onListen`, but rather we need to make sure we're throwing an error at the right point in the ST `onListen` callback function. I think it might be anough to

  > I do wish we had better options regarding Dart Streams, a lot of its inner workings are private implementations and just the way they work is counter intuitive to what other Stream implementations do, but that's a different story ;)

Brother, tell me about it :) There have been a few times where I wondered if I should submit PRs making some of this stuff accessible, haha.

frankpepermans:
@brianegan Indeed, it throws too late, but this is a larger issue, for example:
```
test('doOnListen accidental broadcast', () async {
    final controller = StreamController<int>();

    final stream = controller.stream.switchMap((_) {
      return Stream<int>.value(1);
    });

    stream.listen(null);
    expect(() => stream.listen(null), throwsStateError);

    controller.add(1);
  });
```

has the same issue

PR #396, opened 2020-01-25
Add distinctBy

yongjhih:
```dart
Stream.fromIterable(['123', 'abc']).distinctBy((value) => value.length),
        emitsInOrder(<dynamic>[
          '123',
          emitsDone
        ]));
```

### DIFF
--- a/lib/src/transformers/distinct_by.dart
+++ b/lib/src/transformers/distinct_by.dart
@@ -1,0 +1,14 @@
+import 'dart:async';
+
+/// Extends the Stream class with the ability to skip items that have previously
+/// been emitted.
+extension DistinctByExtension<T> on Stream<T> {
+  /// Creates a Stream where data events are skipped if they have already
+  /// been emitted before.
+  ///
+  /// [Interactive marble diagram](http://rxmarbles.com/#distinct)
+  Stream<T> distinctBy<S>(
+    S Function(T it) distincter,
+  ) =>
+      distinct((prev, next) => distincter(prev) == distincter(next));
+}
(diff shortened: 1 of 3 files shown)


frankpepermans:
Heya,

Thanks for submitting a PR!

However, I think it's best if we try to keep the amount of operators to a minimum,
`distinctBy(method)` is only a shorthand to `distinct((a, b) => method(a) == method(b))`.

You could do similar things for all other operators too, to suit different special use cases, but in the end, the sheer amount of operators would be too overwhelming even to experienced users.

Plus, you could simply enable this extension in your own project easily.

yongjhih:
In RxJava and RxJs, the `distinct` operator is a selector instead of a comparator, that's inconsistent. IMO, we should provide an equivalent method.

According to you mention that keep the amount of operators to a minimum. I think the `mapTo()` redundant operator, because it's just a shorthand to `map((it) => true)`.

frankpepermans:
Remember, rxdart is built on top of Dart Streams.

In Dart Streams, there's distinct. But with different behavior to Rx in general, where we have distinct and distinctUntilChanged.

So, since distinct is taken, we chose to add distinctUnique and gave it the same signature, i.e. an (a, b) => bool.

The above is confusing to begin with, if we then add alternatives using selectors, we'd get:
distinct, distinctBy, distinctUnique, distinctUniqueBy.

...which is my main concern here.

Maybe selectors in general need a separate issue, cos we'd want to be consistent throughout the whole library then.

Just a random thought, but we could make selector alternatives available in a separate lib, ie rxdart/selectors.dart which is an optional import.

That way the common operators would follow Dart standards, where the new import would follow classic Rx standards.

hoc081098:
For anyone interested in this extension, consider my package https://github.com/hoc081098/rxdart_ext 😃

PR #397, opened 2020-01-28
Add error and hasError to BehaviorSubject

najeira:
This fix allows to get the last added error and hasError, like value and hasValue.

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -170,6 +170,12 @@ class ValueConnectableStream<T> extends ConnectableStream<T>
 
   @override
   bool get hasValue => _subject.hasValue;
+
+  @override
+  Object get error => _subject.error;
+
+  @override
+  bool get hasError => _subject.hasError;
 }
 
 /// A [ConnectableStream] that converts a single-subscription Stream into
--- a/lib/src/streams/value_stream.dart
+++ b/lib/src/streams/value_stream.dart
@@ -6,4 +6,11 @@ abstract class ValueStream<T> implements Stream<T> {
 
   /// A flag that turns true as soon as at least one event has been emitted.
   bool get hasValue;
+
+  /// Last emitted error, or null if no error added or value exists.
+  /// See [hasError]
+  Object get error;
+
+  /// A flag that turns true as soon as at an error event has been emitted.
+  bool get hasError;
 }
--- a/lib/src/subjects/behavior_subject.dart
+++ b/lib/src/subjects/behavior_subject.dart
@@ -133,6 +133,12 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
 
   /// Set and emit the new value
   set value(T newValue) => add(newValue);
+
+  /// A flag that turns true as soon as at an error event has been emitted.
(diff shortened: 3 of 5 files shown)


najeira:
Should error and hasError be added to ValueStream?

brianegan:
Heya @najeira -- Thanks for contributing, and sorry for the delay. 

Dang, I always feel like this is a bit of a weird one to add, but you're not the first to ask for it. In the context of Dart Streams being able to emit multiple errors, I think it probably makes sense. @frankpepermans -- as a sanity check, what do you think?

And, yes, you're right it make sense to add this to the `ValueStream` interface as well.

Thanks again!

frankpepermans:
Seems like a good addition, but I'm a bit worried about the api here:
we would have:
- hasValue
- value
- hasError
- error

and that's simply confusing:
If `hasError` and error are filled, do you expect `hasValue` and value to be empty?

if the last event was an Error, then I'd say no, 
but someone else might expect just the latest value here, the last emitted event as Error doesn't impact that.

also, figuring out what the last event exactly was is non-trivial:
```
if (subject.hasError || subject.hasValue) {
    // yup, there was an event alright
}

if (subject.hasValue) {
    // wait, maybe it did emit a value, but also errored?
    if (subject.hasError) {
        // ok, but what was the very last event then, value or error?
    }
}
```
Say we do clear value and `hasValue` whenever a last event is an Error, any code checking on `hasValue` becomes non-deterministic. As a value may happen, but later that same value becomes `null`.

Maybe we could do something else, which resembles the BloC `state` property more?

```
abstract class ValueSubjectState<T> {
   // no events yet, or errors
   bool get isIdle; 
   // last event is of T
   bool get latestIsValue;
   // last event was an Error
   bool get latestIsError; 

   // last known Error emitted
   Object get lastKnownError;
   // last known event T emitted
   T get lastKnownValue;
}
```
This way, imagine a Subject with emits [value1, value2, error1, value3]
then, the above object would be:

`ValueSubjectState(isIdle: false, latestIsValue: true, latestIsError: false, lastKnownError: error1, lastKnownValue: value3)`

najeira:
I think it is reasonable that the value is empty when the error is emitted. But I think some people don't. We need to add document for this.

​​ValueSubjectState is also looks good.

Nico04:
Any update on this ?

najeira:
No updates...

This PR does not break existing code. Only `error` property has been added, there is no change in the behavior of `value` property. So I want this to be merged.

@frankpepermans  's idea of `ValueSubjectState` and `subject.state` seems to be good, but it breaks compatibility and should be discussed in another Issue or PR.

How should this PR be handled?

CC: @brianegan

frankpepermans:
@najeira sorry been very busy throughout the lockdown:
I still feel it could be handled better, but then again, especially with Flutter, it would really help a lot of people.
We could always revisit this later as well...

Ok with the PR as is,
dunno where @brianegan is, I'll wait a few more days before merging :)

TercyoStorck:
I guess we don't need error value, but a hasError property will be very useful!

cassioseffrin:
any update on this?

Nico04:
For me it's available since 0.25.0 👍

Issue #404, opened 2020-02-26
Subjects throw exception when an error is piped in, even if it is handled


ferdkuh:
It seems that when an error is piped into any of the `Subject` subclasses, an exception is thrown even if it is handled, either by an `handleError` or an `onError` callback in `listen`.

When adding errors directly to the subject or using a `StreamController`, this does not happen, so I am guessing that this is not intended behaviour. I am using the latest (0.23.1) version of RxDart.

Here's a minimal example to reproduce the issue:

```dart
import 'dart:async';
import 'package:rxdart/rxdart.dart';

void main() {
  // works as expected with StreamController but
  // throws an exception with any Subject subclass

  final source = StreamController<int>();
  final target = BehaviorSubject<int>();
//  final target = StreamController<int>();

  source.stream.pipe(target);
  target.stream
      // handleError is called, but exception is thrown nevertheless
      .handleError((err, _) => print("handleError: $err"))
      .listen((n) => print("n = $n"));

  source.addError("oh oh");
}
```

hoc081098:
The exception are not thrown by Stream, but rather the exception is thrown by Future retuned from pipe method [subject.dart line 112](https://github.com/ReactiveX/rxdart/blob/ff0a6d7dbcf2c7bd4c375bad19b43df127fd6466/lib/src/subjects/subject.dart#L112)

PR #405, opened 2020-03-02
bug/404

frankpepermans:
Fixes https://github.com/ReactiveX/rxdart/issues/404

### DIFF
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -100,6 +100,14 @@ abstract class Subject<T> extends StreamView<T> implements StreamController<T> {
     }
 
     final completer = Completer<T>();
+    var isOnDoneCalled = false;
+    final complete = () {
+      if (!isOnDoneCalled) {
+        isOnDoneCalled = true;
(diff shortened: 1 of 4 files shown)


lib/src/subjects/subject.dart
```diff
@@ -108,12 +116,10 @@ abstract class Subject<T> extends StreamView<T> implements StreamController<T> {
       _controller.addError(e, s);
 
       if (cancelOnError) {
-        _isAddingStreamItems = false;
-        completer.completeError(e);
```
brianegan: Just curious: Why would ya want to complete instead of completing with an error here? Guessing it's related to the cause of the bug, was wondering about it :)

frankpepermans:
@brianegan Yes, tests fail, but only in Travis it seems, locally everything succeeds :/

frankpepermans:
@brianegan tests up again, merged the fixes into this branch

lib/src/subjects/subject.dart
brianegan: Overall, PR looks good, but was still curious about this part :)

PR #408, opened 2020-03-03
fix tests

frankpepermans:
To check:

Should we just transform to a broadcast `Stream`?
Or place an assert checking `isBroadcast` instead?

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -101,7 +101,10 @@ class ValueConnectableStream<T> extends ConnectableStream<T>
   final Stream<T> _source;
   final BehaviorSubject<T> _subject;
 
-  ValueConnectableStream._(this._source, this._subject) : super(_subject);
+  ValueConnectableStream._(Stream<T> source, this._subject)
+      : _source =
+            source.isBroadcast ?? true ? source : source.asBroadcastStream(),
+        super(_subject);
 
   /// Constructs a [Stream] which only begins emitting events when
   /// the [connect] method is called, this [Stream] acts like a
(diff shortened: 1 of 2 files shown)


lib/src/transformers/exhaust_map.dart
```diff
@@ -74,7 +74,7 @@ class ExhaustMapStreamTransformer<T, S> extends StreamTransformerBase<T, S> {
         },
         onCancel: () async {
           await inputSubscription.cancel();
-          if (outputIsEmitting) await outputSubscription.cancel();
+          if (outputIsEmitting) await outputSubscription?.cancel();
```
brianegan: Do we know what made this start failing? This test literally isn't failing on my machine, and I'm just a bit worried something funky is going on with Travis even on the stable version of Dart + Master branch of RxDart. 

I always worry about adding null checks in like this unless we're really sure it isn't needed, otherwise this type of defensive programming can hide the actual bug.

frankpepermans: It is on mine and on Travis, I'm just fixing up in this PR, since outputIsEmitting is true here, then outputSubscription cannot be null, I'm baffled why this happens atm :)

brianegan:
Nice one! Just took a look independently and came to the same conclusion that this is a goo better way to solve it rather than the null check. 

I'm also baffled as to why it used to work -- the old code definitely looks wrong, haha. How would it create the subscription at all if the function threw an error?

lib/src/streams/connectable_stream.dart
```diff
@@ -49,7 +49,10 @@ class PublishConnectableStream<T> extends ConnectableStream<T> {
     return PublishConnectableStream<T>._(source, PublishSubject<T>());
   }
 
-  PublishConnectableStream._(this._source, this._subject) : super(_subject);
+  PublishConnectableStream._(Stream<T> source, this._subject)
+      : _source =
+            source.isBroadcast ?? true ? source : source.asBroadcastStream(),
```
brianegan: Does `isBroadcast` return null in some cases?

PR #414, opened 2020-03-12
adds end_with and end_with_many

frankpepermans:
Had a few use cases for this, at least RxJs supports it too.


lib/src/transformers/end_with_error.dart
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/// Appends an error to the source [Stream] before closing.
+///
+/// ### Example
+///
+///     Stream.fromIterable([2])
+///       .transform(EndWithErrorStreamTransformer('error'))
+///       .listen(null, onError: (e) => print(e)); // prints 'error'
+class EndWithErrorStreamTransformer<S> extends StreamTransformerBase<S, S> {
```
brianegan: I still don't know about these ones... feels like a bad practice to encourage folks to emit errors. What do you think of leaving this out unless someone requests it, or did you need it in particular?

test/rxdart_test.dart
```diff
@@ -51,6 +51,8 @@ import 'transformers/dematerialize_test.dart' as dematerialize_test;
 import 'transformers/distinct_test.dart' as distinct_test;
 import 'transformers/distinct_unique_test.dart' as distinct_unique_test;
 import 'transformers/do_test.dart' as do_test;
+import 'transformers/end_with_test.dart' as end_with_test;
```
brianegan: If we're including error, should be here

lib/transformers.dart
```diff
@@ -5,6 +5,8 @@ export 'src/transformers/delay.dart';
 export 'src/transformers/dematerialize.dart';
 export 'src/transformers/distinct_unique.dart';
 export 'src/transformers/do.dart';
+export 'src/transformers/end_with.dart';
```
brianegan: If we're including error, should be here

lib/src/transformers/end_with_error.dart
frankpepermans: This is actually by mistake, we have StartWithError, which is private and only used with BehaviorSubject/ReplaySubject.

But EndsWithError has no purpose, I simply copied all from start_with_x and did changes :)

Issue #419, opened 2020-03-22
mapping a value stream to a new value stream synchronously


khainke:
i don't know whether this is even feasible, but ideally I would like to have the ability to map a value stream in such a way to a new value stream, that the last value is instantly mapped to the created stream and not after a delay. Example code:

```dart
test('shareValue does not work synchronously', () async {
      BehaviorSubject<int> controller = BehaviorSubject.seeded(1);
      ValueStream<int> stream = controller.stream;

      int plusOne(int i) => i + 1;

      ValueStream<int> streamPlusOneNaive = stream.map(plusOne).shareValue();
      ValueStream<int> streamPlusOneSeeded = stream.hasValue
          ? stream.map(plusOne).shareValueSeeded(plusOne(stream.value))
          : stream.map(plusOne).shareValue();

      StreamSubscription<int> streamPlusOneNaiveSubscriber = streamPlusOneNaive.listen((_){});
      StreamSubscription<int> streamPlusOneSeededSubscriber = streamPlusOneSeeded.listen((_){});

      // ideally value would be available here without manual seeding it
      expect(stream.value, equals(1));
      expect(streamPlusOneNaive.value, isNull);
      expect(streamPlusOneSeeded.value, equals(2));

      await Future.delayed(Duration(milliseconds: 5));

      expect(stream.value, equals(1));
      expect(streamPlusOneNaive.value, equals(2));
      expect(streamPlusOneSeeded.value, equals(2));

      controller.add(2);

      // ideally values would be mapped here already
      expect(stream.value, equals(2));
      expect(streamPlusOneNaive.value, equals(2));
      expect(streamPlusOneSeeded.value, equals(2));

      await Future.delayed(Duration(milliseconds: 5));

      expect(stream.value, equals(2));
      expect(streamPlusOneNaive.value, equals(3));
      expect(streamPlusOneSeeded.value, equals(3));

      streamPlusOneNaiveSubscriber.cancel();
      streamPlusOneSeededSubscriber.cancel();
    });
```

frankpepermans:
Not easy I'm afraid,
`refCount` is an extension method on `Stream`, and we have no way of knowing if a `Stream` is either sync or async...

When using `refCount`, we switch to a `BehaviorSubject` underneath obviously, but given that we don't know if the `Stream` is sync or async, we've chosen to always make the underlying controller async by default.

Ideally, this should fix things the way you'd want it, without a seeded hack:
[...]

PR #420, opened 2020-03-23
Sync Streams behavior

frankpepermans:
See issue 419: https://github.com/ReactiveX/rxdart/issues/419

...which is interesting,
sure we could argue that `Stream`s are async by design, and that expecting values to be available sync is not guaranteed, but *we do have sync Streams in Dart obviously*...

So, I took the same test as within that issue and made it expect sync instead of async:
```
test('shareValue does not work synchronously', () async {
    BehaviorSubject<int> controller = BehaviorSubject.seeded(1, sync: true);
    ValueStream<int> stream = controller.stream;

    int plusOne(int i) => i + 1;

    ValueStream<int> streamPlusOneNaive = stream.map(plusOne).shareValue();
    ValueStream<int> streamPlusOneSeeded = stream.hasValue
        ? stream.map(plusOne).shareValueSeeded(plusOne(stream.value))
        : stream.map(plusOne).shareValue();

    StreamSubscription<int> streamPlusOneNaiveSubscriber = streamPlusOneNaive.listen((_){});
    StreamSubscription<int> streamPlusOneSeededSubscriber = streamPlusOneSeeded.listen((_){});

    print('is: ${streamPlusOneNaive.value}');
    // ideally value would be available here without manual seeding it
    expect(stream.value, equals(1));
    expect(streamPlusOneNaive.value, equals(2));
    expect(streamPlusOneSeeded.value, equals(2));

    await Future<void>.delayed(Duration(milliseconds: 5));

    expect(stream.value, equals(1));
    expect(streamPlusOneNaive.value, equals(2));
    expect(streamPlusOneSeeded.value, equals(2));

    controller.add(2);

    // ideally values would be mapped here already
    expect(stream.value, equals(2));
    expect(streamPlusOneNaive.value, equals(3));
    expect(streamPlusOneSeeded.value, equals(3));

    await Future<void>.delayed(Duration(milliseconds: 5));

    expect(stream.value, equals(2));
    expect(streamPlusOneNaive.value, equals(3));
    expect(streamPlusOneSeeded.value, equals(3));

    streamPlusOneNaiveSubscriber.cancel();
    streamPlusOneSeededSubscriber.cancel();
  });
```

...which works in this PR, but:

- I use a special version of `startWith`/`startWithError` to make it work
- Only when the *source* is sync, then add the very first event (or error) sync as well

This then works, note that immediately emitting for async controllers does throw the RTE.
Following the rabbit hole of the stack trace, at some point in the SDK a `subscription` is created, with an `onListen` handler. This then in its turn does something when data is added.
Now when everything is sync, then this chain fires immediatly, and at some point, that subscription is being referenced. (which yields the RTE, since at this point, it is effectively null).

TL;DR that subscription is lazy (or late in Dart NNBD terms), but unfortunately lazy null-safety can still yield NPE's, as in this case.

TL;DR2 I do think that we should support true sync behavior, as expected by the test above, but I'm not too sure yet about the implementation here (just a PoC at this point).

Note: Yes, this PR also implies that Subjects created via share are sync, but so are all of our Transformers, again this won't magically turn the source into a sync one by doing this:
The source still emits async, but any effects afterwards run in sync, so in total, it is still async.

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -45,8 +45,8 @@ class PublishConnectableStream<T> extends ConnectableStream<T> {
   /// Constructs a [Stream] which only begins emitting events when
   /// the [connect] method is called, this [Stream] acts like a
   /// [PublishSubject].
-  factory PublishConnectableStream(Stream<T> source) {
-    return PublishConnectableStream<T>._(source, PublishSubject<T>(sync: true));
+  factory PublishConnectableStream(Stream<T> source, {bool sync = false}) {
+    return PublishConnectableStream<T>._(source, PublishSubject<T>(sync: sync));
   }
 
   PublishConnectableStream._(Stream<T> source, this._subject)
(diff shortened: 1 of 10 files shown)


frankpepermans:
@brianegan now no longer fails <- was due to the other PR about test flakyness

brianegan:
Cool, looks good overall!

  > TL;DR that subscription is lazy (or late in Dart NNBD terms), but unfortunately lazy null-safety can still yield NPE's, as in this case.

THIS IS SUCH A PAIN! I've tried to hack my way around this exact issue so many times. If the SDK didn't quite do this we could actually do some really cool stuff with Sync streams.

PR #426, opened 2020-04-05
Add IntervalStream

### LINKED ISSUE #225: Interval operator
I'm used to working with RxJava and I'm missing the [interval](http://reactivex.io/RxJava/javadoc/io/reactivex/Flowable.html#interval-long-long-java.util.concurrent.TimeUnit-) operator.
As a replacement I tried `Observable.periodic`, but it has no `initialDelay` argument.

Any workarounds or possibility of a feature request?

Thanks

Wheater:
* add new IntervalStream
* add IntervalStream tests
* Fix #225


brianegan:
Heya @Wheater -- thanks so much for the contribution! One big question for me: The Stream class already contains the `Stream.periodic` constructor, which works in almost the exact same way.

Could you describe the benefit of including this class in addition to the good ol' `Stream.periodic`?

Wheater:
@brianegan Honestly, I was going off the recommended issue list for those getting started. Now that you pose the question, it provides some familiarity for devs coming from RxJS. It is also part of the ReactiveX documentation, so having a stream for it provides a familiar interface for those expecting it exists. 

If you don't think it's necessary, I can see that side as well given that Stream.periodic is so similar.

frankpepermans:
If we provide IntervalStream, equivalent to Stream.periodic, then IMO it's more confusing for users.
In this case, it's better to add docs for peeps who are looking for IntervalStream instead.

https://github.com/ReactiveX/rxdart/issues/225 is the original issue, which *might* warrant an rxdart entry, although you could easily do:

```
Stream<T>interval(Duration initialDelay, Duration nextDelay, T Function(int index) compute) async* {
    yield await Future.delayed(initialDelay).then(0).then(compute);

    yield* Stream.periodic(nextDelay, (index) => index + 1).map(compute);
}
```

Wheater:
Sounds good, I will work on updating the docs for that instead.

brianegan:
Thanks again, @Wheater! Really appreciate your efforts to dive in and help out :)

PR #429, opened 2020-04-05
🐛 Fix RetryWhenStream.cancel sub stream cancellation

### LINKED ISSUE #428: RetryWhenStream.cancel does not cancel the sub stream
Hello 👋

I tried to implement sort of exponential backoff retrying using `RetryWhenStream`.
It works pretty well, but I encountered a bug when trying to cancel this stream, it runs indefinitely and ignores my cancellation.
After some investigations, it seems that we can't cancel the sub stream by cancelling the main stream.

Here is an example to reproduce this bug:
```dart
import 'dart:async';

import 'rxdart.dart';

void main() async {
  StreamSubscription<int> subscription;

  var retries = 5;

  subscription = RetryWhenStream(
    _sourceStream(3, 0),
    (dynamic _, dynamic __) {
      print('retry $retries');
      if (retries == 0) return Stream.error('error');
      retries--;
      return TimerStream(0, Duration(milliseconds: 100));
    },
  ).listen((_) {});

  await Future<dynamic>.delayed(Duration(milliseconds: 10));
  await subscription.cancel();
}

Stream<int> Function() _sourceStream(int i, [int throwAt]) {
  return throwAt == null
      ? () => Stream.fromIterable(range(i))
      : () =>
          Stream.fromIterable(range(i)).map((i) => i == throwAt ? throw i : i);
}

List<int> range(int i) => List<int>.generate(i, (i) => i);
```

This code outputs:
```
$ dart lib/retry_bug.dart
retry 5
retry 4
retry 3
retry 2
retry 1
retry 0
```

It should stop at the first iteration, but it continues until the end.
I don't really know all the internally rxdart stack, but with what I can see, the `StreamController` in `lib/src/streams/retry_when.dart` should cancel the sub stream with the main stream.

I am going to link a pull request to try to fix this.
After applying the pull request, the output of the same code is:
```
$ dart lib/retry_bug.dart
retry 5
```

ThomasKliszowski:
Fix #428

### DIFF
--- a/lib/src/transformers/backpressure/throttle.dart
+++ b/lib/src/transformers/backpressure/throttle.dart
@@ -52,7 +52,7 @@ extension ThrottleExtensions<T> on Stream<T> {
   /// ### Example
   ///
   ///     Stream.fromIterable([1, 2, 3])
-  ///       .throttle((_) => TimerStream(true, Duration(seconds: 1)))
+  ///       .throttle((_) => TimerStream(true, Duration(seconds: 1)));
   Stream<T> throttle(Stream Function(T event) window,
           {bool trailing = false, bool leading = true}) =>
       transform(
@@ -72,7 +72,7 @@ extension ThrottleExtensions<T> on Stream<T> {
   /// ### Example
   ///
   ///     Stream.fromIterable([1, 2, 3])
-  ///       .throttleTime(Duration(seconds: 1))
+  ///       .throttleTime(Duration(seconds: 1));
   Stream<T> throttleTime(Duration duration,
       {bool trailing = false, bool leading = true}) {
     ArgumentError.checkNotNull(duration, 'duration');
--- a/lib/src/transformers/do.dart
+++ b/lib/src/transformers/do.dart
@@ -194,7 +194,7 @@ extension DoExtensions<T> on Stream<T> {
   /// ### Example
   ///
   ///     final subscription = TimerStream(1, Duration(minutes: 1))
-  ///       .doOnCancel(() => print('hi'));
+  ///       .doOnCancel(() => print('hi'))
   ///       .listen(null);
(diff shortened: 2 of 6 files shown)


ThomasKliszowski:
👋 @brianegan 
Yes indeed, a test case would be great to prevent regressions.
I looked at the existing RetryWhenStream test cases, but I don't really know how to test that the cancel is not working as it's all asynchronous.
The TimerStream approach does not seem to be the good one, timers in test cases slow down the process.

How would you do it?

frankpepermans:
@ThomasKliszowski Not the prettiest of tests, but it does the trick:

```
test('RetryWhenStream.ensure sub stream cancels', () async {
    var isCancelled = false, didStopEmitting = true;
    final subStream = (dynamic e, StackTrace s) =>
        Stream.periodic(const Duration(milliseconds: 100), (count) => count)
            .doOnData((_) {
          if (isCancelled) {
            didStopEmitting = false;
          }
        });
    final subscription =
        RetryWhenStream(_sourceStream(3, 0), subStream).listen(null);

    await Future<void>.delayed(const Duration(milliseconds: 250));

    await subscription.cancel();
    isCancelled = true;

    await Future<void>.delayed(const Duration(milliseconds: 250));

    expect(didStopEmitting, isTrue);
  });
```

ThomasKliszowski:
Sorry about the delay guys, it totally slipped my mind.
Thanks @frankpepermans for the test, it seems there is no better way than dealing with timers then, fair enough.

Here you go, tell me if that works for you!

ThomasKliszowski:
Hi @brianegan @frankpepermans , do you mind reviewing this one?

I keep using my fork for the time being, it could be easier to switch back to the main repo.

Thank you guys 🙏

PR #432, opened 2020-04-07
Add addTo extension method for StreamSubscription

jarekb123:
addTo extension method - adds subscription to CompositeSubscription

Example
```dart
final subscriptions = CompositeSubscription();

stream
   .listen((data) => print(data))
   .addTo(subscriptions)
```

### DIFF
--- a/lib/src/utils/composite_subscription.dart
+++ b/lib/src/utils/composite_subscription.dart
@@ -55,3 +55,10 @@ class CompositeSubscription {
     _isDisposed = true;
   }
 }
+
+/// Extends the [StreamSubscription] class with the ability to be added to [CompositeSubscription] container.
+extension AddToCompositeSubscriptionExtension<T> on StreamSubscription<T> {
+  /// Adds this subscription to composite container for subscriptions.
+  void addTo(CompositeSubscription compositeSubscription) =>
+      compositeSubscription.add(this);
+}


Wheater:
Wow, that looks very nice!

frankpepermans:
I like the idea, just thinking out loud, for discoverability, how about

```
final subscriptions = CompositeSubscription();

stream
   .listenManaged((data) => print(data), compositeSubscription: subscriptions);
```

frankpepermans:
...or
compositeSubscription.listen(stream, onData: ...)

brianegan:
I like it! @frankpepermans -- sorry, I'm a bit confused about your suggestions. The `CompositeSubscription` class already has the `add` method which takes in the StreamSubscription, so I don't think it would simplify things to change that up.

I think it's kind of a nice and simple extension that fits well with Rx.

jarekb123:
The syntax comes from RxKotlin.

https://github.com/ReactiveX/RxKotlin/blob/3.x/src/main/kotlin/io/reactivex/rxjava3/kotlin/disposable.kt

frankpepermans:
Oh, I actually meant to change the `CompositeSubscription` to have `addStream` and `addSubscription`.

For the record, I like this PR too, was just bringing up possible alternatives.

PR #433, opened 2020-04-08
Fixes withLatestFrom

frankpepermans:
I've noticed 2 issues with our current `withLatestFrom`:
- ~~it closes too early, it should close when: the source closes and all combined Streams close~~
- we kept a reference to a single Subscription only when listening to the combined Steams, now we keep a List and properly handle `cancel()` as well.

### DIFF
--- a/lib/src/transformers/with_latest_from.dart
+++ b/lib/src/transformers/with_latest_from.dart
@@ -9,7 +9,9 @@ class _WithLatestFromStreamSink<S, T, R> implements ForwardingSink<S> {
   final EventSink<R> _outputSink;
   final List<bool> _hasValues;
   final List<T> _latestValues;
-  StreamSubscription<T> _subscription;
+  List<StreamSubscription<T>> _subscriptions;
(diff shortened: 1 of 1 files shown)


hoc081098:
- Issue 1: In [RxJava-ObservableWithLatestFromMany](https://github.com/ReactiveX/RxJava/blob/b47783e3186f83be6a87257a2d13c9619bccaa42/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableWithLatestFromMany.java#L194) and [RxSwift-WithLatestFrom](https://github.com/ReactiveX/RxSwift/blob/29e484f264ef365f13dc472a1b20ce51f1a1c0a4/RxSwift/Observables/WithLatestFrom.swift#L84), result `Observable` closes as soons as source `Observable` is closed
- Issue 2:

frankpepermans:
@hoc081098 you're right!
removed that change

hoc081098:
@frankpepermans Thanks

PR #441, opened 2020-04-13
Fix concatWith on broadcast stream

jarekb123:
concatWith on broadcast stream returns multi subscription stream


frankpepermans:
thanks!

PR #442, opened 2020-04-13
fix onCancel being called right after close

frankpepermans:
When using a `ForwardingStream`, when the inner `StreamController` closes, then `close` and `onCancel` are called immediately.

For some transformers, like `flatMap`, `switchMap`, ... we want to keep listening after the main `Stream` closes, and only call `onCancel` when all underlying subscriptions cancel.

`safeClose` is therefore added, returning a `Future`.
If a `Future` is returned, then the main controller will await it before actually closing.

### DIFF
--- a/lib/src/transformers/exhaust_map.dart
+++ b/lib/src/transformers/exhaust_map.dart
@@ -8,6 +8,7 @@ class _ExhaustMapStreamSink<S, T> implements ForwardingSink<S> {
   final EventSink<T> _outputSink;
   StreamSubscription<T> _mapperSubscription;
   bool _inputClosed = false;
+  Completer _canClose;
 
   _ExhaustMapStreamSink(this._outputSink, this._mapper);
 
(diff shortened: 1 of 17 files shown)


lib/src/transformers/exhaust_map.dart
```diff
@@ -19,26 +20,40 @@ class _ExhaustMapStreamSink<S, T> implements ForwardingSink<S> {
...
+          }
+        });
   }
 
   @override
   void addError(e, [st]) => _outputSink.addError(e, st);
 
   @override
-  void close() {
+  Future safeClose() {
```
brianegan, line 42: Should this be a `FutureOr` instead of Future? Or is it more a `Future|Null` union?

frankpepermans: Future|null indeed, if null, then the parent controller will just close in sync.
This would become a Future? if nnbd drops

brianegan: Sounds good!

PR #446, opened 2020-04-15
Fix: BehaviorSubject replays a data event instead of error for new subscriptions when using addStream

### LINKED ISSUE #445: BehaviorSubject replays a data event instead of error for new subscriptions when using addStream
When BehaviorSubject.addStream is used to feed a subject from another stream with cancelOnError=false, if the last event was an error, new subscribers will not get this error upon subscription. Instead they will be fed with the last data event.

yuvalr1:
Fixes #445

### DIFF
--- a/lib/src/subjects/subject.dart
+++ b/lib/src/subjects/subject.dart
@@ -113,7 +113,7 @@ abstract class Subject<T> extends StreamView<T> implements StreamController<T> {
     source.listen((T event) {
       _add(event);
     }, onError: (dynamic e, StackTrace s) {
-      _controller.addError(e, s);
+      _addError(e, s);
 
       if (cancelOnError) {
         complete();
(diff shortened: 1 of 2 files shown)


yuvalr1:
To elaborate a little bit:
The root cause for the bug is that the Subject parent doesn't call addError from addStream. The BehaviorSubject child counts on it to update the state of its latestIsError variable (in the wrapper). 

Because the variable stays false even on error, a data event is emitted instead.

The commit carries the word "null data event", but instead it's actually "the last data event". In my case it was null because of the way I reproduced it.

frankpepermans:
@yuvalr1 thanks!

@brianegan you think it's safe to call `_addError` directly?
I see that the public `addError` throws when `addStream` is active, so we're now bypassing this...

brianegan:
Ah, good call -- yes, we'd need to use the private `_addError` here, because we trigger the `_isAddingStreamItems` is set at the top of this function. Therefore, if we ever called the public `addError` from this context, it would be guaranteed to fail since it checks for that exact boolean.

PR #449, opened 2020-04-19
refactor(min, max): rewrite min, max operator without sorting

hoc081098:
Rewrite min, max operator without sorting list

### DIFF
--- a/lib/src/transformers/max.dart
+++ b/lib/src/transformers/max.dart
@@ -1,3 +1,5 @@
+import 'package:rxdart/src/utils/min_max.dart';
+
 /// Extends the Stream class with the ability to transform into a Future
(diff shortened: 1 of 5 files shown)


frankpepermans:
I was thinking we could optimise it better though, if a large amount of say, 1 million events is being added, then we still have a performance hit onDone, with the old sort or the reduce loop.

Maybe we can just subscribe instead, and on every added event run the compare, and keep the current min/max, which repeats on the next emitted event?

hoc081098:
I removed `toList`, replaced by listening stream and repeatedly applying `comparator`

frankpepermans:
Looks good!

brianegan:
Woohoo! Great rewrite :)

PR #450, opened 2020-04-19
Refactor stream transformer

hoc081098:
* Rewriting forward stream transformer, now doesn't need `SafeClose` interface to fix close behavior
* Cancel subscriptions of other streams when source stream is cancelled
* Fix some typos in `CompositeSubscription`
* Using `Stream.value(value)` instead of `Stream.fromIterables([value])` in `OnErrorExtensions`
* `_TakeWhileInclusiveStreamSink` implements `EventSink` instead of `ForwardingSink`

### DIFF
--- a/lib/src/subjects/behavior_subject.dart
+++ b/lib/src/subjects/behavior_subject.dart
@@ -135,7 +135,7 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
   set value(T newValue) => add(newValue);
 
   @override
-  StreamController<T> createForwardingController({
+  StreamController<R> createForwardingController<R>({
     void Function() onListen,
     void Function() onCancel,
     bool sync = false,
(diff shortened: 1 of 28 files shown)


frankpepermans:
That's a big one :)

I guess the biggest change would be, instead of using Stream.eventTransformed and providing a Sink, you're opted to calling that same sink, but directly from the controller, and returning its Stream?

Stream.eventTransformed kind of does this behind the scenes, right now, I guess my only concern would be that if they decide to optimise the inner workings behind the scenes, we'd lose out on that.

Other than that, it does simplify things.

Thanks!

frankpepermans:
I've been able to try out this PR in a large project now, looks good so far.
Gonna do this for a few days, if nothing pops up, then ok to merge for me.

frankpepermans:
Couldn't spot any issues after almost a week of running it in a very large app,
merged!

hoc081098:
Oh, I'm glad for that 👍

PR #454, opened 2020-04-30
update(subject): remove Subject.createForwardingController

hoc081098:
remove `Subject.createForwardingController`, as it is not in use


frankpepermans:
Yup, good riddance!

PR #457, opened 2020-05-05
Rx.combineLatestList([]) emits a single empty result.

jackdreilly:
* Before this change: `Rx.combineLatestList([])` never emits, and never completes.
* After this change: it emits a single empty list, and then completes.

The latter behavior is consistent with RxSwift https://github.com/ReactiveX/RxSwift/commit/5289de9a1cb531b36733db2601c925ee23a554e4

The behavior of rxjs is to complete immediately and emit nothing, but this seems less useful and counter-intuitive: https://rxjs-dev.firebaseapp.com/api/index/function/combineLatest

### DIFF
--- a/lib/src/streams/combine_latest.dart
+++ b/lib/src/streams/combine_latest.dart
@@ -287,6 +287,11 @@ class CombineLatestStream<T, R> extends StreamView<R> {
     Iterable<Stream<T>> streams,
     R Function(List<T> values) combiner,
   ) {
+    if (streams.isEmpty) {
+      return StreamController()
+        ..add(combiner([]))
+        ..close();
+    }
     final len = streams.length;
     List<StreamSubscription<dynamic>> subscriptions;
     StreamController<R> controller;
--- a/test/streams/combine_latest_test.dart
+++ b/test/streams/combine_latest_test.dart
@@ -18,6 +18,11 @@ Stream<bool> get streamC {
 }
 
 void main() {
+  test('Rx.combineLatestList.isEmpty', () {
+    expect(Rx.combineLatestList<dynamic>([]),
+        emitsInOrder(<dynamic>[<dynamic>[]]));
+  });
+
   test('Rx.combineLatestList', () async {
     final combined = Rx.combineLatestList<int>([
       Stream.fromIterable([1, 2, 3]),


frankpepermans:
Heya!

Thanks for the PR, 
However, wouldn't it make more sense to just complete without emitting anything though?

Say we pass a List with empty Streams instead, then it should also just complete and not emit anything, I don't see why an empty list would be different?

jackdreilly:
Hi Frank,

As I understand it, `combineLatest` upholds the contract that it will only emit when "all members have emitted at least once". So your counter-example is very easy to comprehend why it would emit nothing, as it explicitly would violate the contract were it to emit anything, since there are streams present in the argument list which have no events emitted.

so: `Rx.combineLatestList([Stream.empty])` should certainly never emit any events.

My example is different, conceptually:

`Rx.combineLatestList([])`

In this case, is it true that "all member have emitted at least once"? If you ask python's `all` function: `all([])` is `True`.  [Here's more](https://stackoverflow.com/questions/19601802/how-does-all-in-python-work-on-empty-lists)

And since it `[].map(fn)` will always be `[]`, then it makes sense to emit the empty list in this case.

Of course, all of this is very ambiguous and IMO under-defined.

It's understandably a matter of taste which direction one chooses to take here (and of course I very well might be missing some major theoretical point).

One thing though is that the stream should certainly be completed, which, currently, it is not.

frankpepermans:
I fully agree that passing an empty List needs to complete, that's a bug.

However emitting an empty List, I dunno.

CombineLatest can only activate when at least one Stream is present. If none, then it cannot do anything, except completing.
Same for zipList or mergeList imo.

jackdreilly:
I believe your experience trumps mine in this area, so I'm happy to defer to your opinion here. For our use case, it's very easy for us to write an extension to handle our typical use case.

Feel free to close (or tweak to solve the `complete`) problem.

frankpepermans:
Hey, not at all!

It's just my opinion, that's all, let's keep this open so that others can chime in. 

We discuss a lot in issues and PRs here :)

frankpepermans:
Breaking down some more, the example:

> And since it [].map(fn) will always be [], then it makes sense to emit the empty list in this case.
In this case, an empty `List` is being _transformed_, since it is effectively empty, it makes sense that this line would yield an empty `Iterable` as well.

A bit more complex: `[].every((it) => it == true)`
Here, we fold the `List` into a single value, using a `check`.
However, the `List` is empty and it yields `true`.
When analysing an every, we can read it as: _"Does every item inside this List pass the check?"_, since the List is empty, we can answer _"Yes"_, because the check can never run and thus never fail either.

Now going to `combineLatestList`,
This isn't a `transformation` or a `fold`, like above, so we're not obliged to yield a result either.
In fact, we don't even have to think about a result, since the `List` is empty.
Futhermore, think about the following lines:
```
var list = const [];
var stream = Rx.comineLatestList(list);
print(list.isEmpty); // true
print(await Stream.fromIterable(list).isEmpty); // true
print(await stream.isEmpty); // I'd expect true here as well then
```

PR #459, opened 2020-05-06
Streams: deal with empty Iterable of streams

hoc081098:
Related to #457 
* Change for operators: `combineLatest`, `concat`, `concatEager`, `forkJoin`, `merge`, `race`, `zip`.
* Change: when passing an empty `Iterable<Stream>`, the resulting stream completes immediately without emitting any items. This behavior is the same that `rxjs`'s, `RxJava`'s operators.


frankpepermans:
We also have https://github.com/ReactiveX/rxdart/pull/457 up for discussion, I agree with this approach, what do you say @brianegan?

PR #467, opened 2020-06-10
Improves pairwise() docs for first event emission.

feinstein:
Explains that pairwise() won't emit the first event until the second one arrives.

/// The first event won't be emitted until the second one arrives.


hoc081098:
Nice improvement

frankpepermans:
Thanks!

PR #470, opened 2020-07-08
Changing letters that conflict on windows.

LuizFilipeMedeira:
This change corrects the error below.

Error on line 6, column 28: Unexpected character.
  ╷
6 │ - Petrus Nguyá»…n ThÃ¡i Há»

(This happens in the version that is in master on windows.)


frankpepermans:
Thanks!

PR #483, opened 2020-07-30
Throttle: separate leading from trailing

### LINKED ISSUE #482: How to implement api call throttling for typeahead autocomplete
Hello. I'm using the following code to make API requests to implement autocomplete:

```
this.query$.throttleTime(Duration(milliseconds: 200), trailing: true).switchMap((q) => _requestSuggestions(q)).listen((items) {
      this.suggestionItems$.add(items);
    });
```

There is a problem with this implementation: at the beginning, it waits 200ms and makes first requests. 

How to implement the same behavior + make the first request instantly? `trailing: false` will not work because it can skip the last request

hoc081098:
Fix #482 : 
```dart
this.query$
    .throttleTime(Duration(milliseconds: 200), trailing: true, leading: true)
    .switchMap((q) => _requestSuggestions(q))
    .listen((items) {
      this.suggestionItems$.add(items);
    });
```
* Separate leading from trailing.
* Updated backpressure/throttle: `Stream` closes when last window closes.


frankpepermans:
Nice!

I'll run it later today when I have more time,
@brianegan are you still available for reviews?

PR #489, opened 2020-08-16
Add UsingStream

hoc081098:
Add `UsingStream` [ReactiveX-Doc](http://reactivex.io/documentation/operators/using.html)
> The Using operator is a way you can instruct an Observable to create a resource that exists only during the lifespan of the Observable and is disposed of when the Observable terminates.


frankpepermans:
Awesome job!

PR #491, opened 2020-08-16
doOnError: update signature

### LINKED ISSUE #492: doOnError: Closure call with mismatched arguments
- Consider below code:
```dart
void main() {
  Stream<void>.error(Exception())
      .doOnError((Object error) => print('Error: $error'))
      .listen(null);
}
```
- When run, will throws:
```
Closure call with mismatched arguments: function 'main.<anonymous closure>'
Receiver: Closure: (Object) => void
Tried calling: main.<anonymous closure>(Instance of '_Exception', Instance of '_StringStackTrace')
Found: main.<anonymous closure>(Object) => void
```

hoc081098:
Fix #492 
- Change:
```dart
Stream<T> doOnError(Function onError)
``` 
to
```dart
Stream<T> doOnError(void Function(Object, StackTrace) onError)
```
- Because we alway call `_onError` with two arguments:
https://github.com/ReactiveX/rxdart/blob/4dcefadf6f72cc5061f7b9346e93179cfd2515d6/lib/src/transformers/do.dart#L46

### DIFF
--- a/lib/src/transformers/do.dart
+++ b/lib/src/transformers/do.dart
@@ -9,7 +9,7 @@ class _DoStreamSink<S> implements ForwardingSink<S, S> {
   final void Function(S event) _onData;
   final void Function() _onDone;
   final void Function(Notification<S> notification) _onEach;
-  final Function _onError;
+  final void Function(Object, StackTrace) _onError;
   final void Function() _onListen;
   final void Function(Future<dynamic> resumeSignal) _onPause;
   final void Function() _onResume;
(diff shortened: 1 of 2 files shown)


lib/src/transformers/do.dart
```diff
@@ -9,7 +9,7 @@ class _DoStreamSink<S> implements ForwardingSink<S, S> {
   final void Function(S event) _onData;
   final void Function() _onDone;
   final void Function(Notification<S> notification) _onEach;
-  final Function _onError;
+  final void Function(Object, StackTrace) _onError;
```
frankpepermans, line 12: I kind of vaguely remember that we once had `StackTrace` in the signature, but ended up removing it, or was it that we allowed it to be just optional?

i.e. provide `void Function _onError`, so you could assign both `doOnError((e) => ...)`  and `doOnError((e, s) => ...)`
and internally we check the provided handler if it accepts one or two args?

hoc081098: I've opened PR #303 before, but I think it should accept 2 args, if we want `StackTrace`  to be optional, use like this `(error, _) { ... }`

PR #502, opened 2020-09-09
Fix RetryStream example and try-catch DeferStream

### LINKED ISSUE #501: Rx.retry wrong example document
https://github.com/ReactiveX/rxdart/blob/4dcefadf6f72cc5061f7b9346e93179cfd2515d6/lib/src/rx.dart#L799

code version https://github.com/ReactiveX/rxdart/releases/tag/0.24.1

```
  ///     Rx
  ///        .retry(() {
  ///          Stream.value(1).concatWith([Stream.error(Error())]); // hint: should return
  ///        }, 1)
  ///        .listen(print, onError: (e, s) => print(e); // Prints 1, 1, RetryError // hint: missed `)` at the end
```

hoc081098:
Closes #501 
-   Fix `RetryStream` example
-   Catch error thrown from `streamFactor`:
```dart
Rx.defer<int>(() => throw Exception())
```

### DIFF
--- a/lib/src/rx.dart
+++ b/lib/src/rx.dart
@@ -791,14 +791,13 @@ abstract class Rx {
   ///
   /// ### Example
   ///
-  ///     Rx.retry(() { Stream.value(1); })
-  ///         .listen((i) => print(i); // Prints 1
+  ///     Rx.retry(() => Stream.value(1))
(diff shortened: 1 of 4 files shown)


frankpepermans (approved):
Nice one!

PR #510, opened 2020-09-30
Add Rx.fromCallable

### LINKED ISSUE #309: Missing Observable.fromCallable
Hi there! You guys are doing an amazing job with this support. However, coming from RxJava2, I was kinda surprised: why there are no methods like `Observable.fromCallable { ... }`? Maybe I'm getting something wrong, but that's the way to do it in RxJava and that seems like an adequate approach. Is there a way to achieve the same idiom in RxDart?

hoc081098:
- Closes #309 
- Seems #310  discontinued, so I open new PR adding `Rx.fromCallable / FromCallableStream`
- Adding param `reusable: bool` that allows listening to Stream multiple times (like `defer`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,23 +86,24 @@ You can construct the Streams provided by RxDart in two ways. The following exam
 
 #### List of Classes / Static Factories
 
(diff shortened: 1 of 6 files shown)


frankpepermans (approved):
No worries about the implementation,
but do you feel that we still need this?

as mentioned in this PR's docs, we have:
`Stream create() async* { yield callable(); }`

or if it is a Future
`callable().asStream();`

hoc081098:
> or if it is a Future
> `callable().asStream();`

I think `Stream create() async* { yield callable(); }`(1) is little different from `callable().asStream();`(2).
`Callable` in (1) called at subscription-time, otherwise `callable` in (2) called at create creation-time

PR #512, opened 2020-10-03
Override BehaviorSubject's built-in operators.

### LINKED ISSUE #500: BehaviorSubject bugged in 0.24.1 and 0.25.0-beta
Probably the same as #477 but I'm opening a new issue because that one was not reopened.

These tests pass in rxdart 0.23.1:
```dart
test('rxdart #477/#500 - a', () async {
  final a = BehaviorSubject.seeded('a')
    .switchMap((_) => BehaviorSubject.seeded('a'))
    ..listen(print);
  await pumpEventQueue();
  expect(await a.first, 'a'); // this future does not complete in 0.24.1 - this is fixed in 0.25.0-beta
});

test('rxdart #477/#500 - b', () async {
  final b = BehaviorSubject.seeded('b')
    .map((_) => 'b')
    .switchMap((_) => BehaviorSubject.seeded('b'))
    ..listen(print);
  await pumpEventQueue();
  expect(await b.first, 'b'); // this future does not complete in 0.25.0-beta
});
```

hoc081098:
Fix #500. Last effort :)

### DIFF
--- a/lib/src/subjects/behavior_subject.dart
+++ b/lib/src/subjects/behavior_subject.dart
@@ -178,7 +178,7 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
       _forwardBehaviorSubject<T>((s) => s.handleError(onError, test: test));
 
   @override
-  Stream<S> expand<S>(Iterable<S> Function(T element) convert) =>
+  ValueStream<S> expand<S>(Iterable<S> Function(T element) convert) =>
       _forwardBehaviorSubject<S>((s) => s.expand(convert));
 
   @override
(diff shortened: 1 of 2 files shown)


frankpepermans (approved):
I guess this is the only solution we have,
nice work!

We should do the same for our ReplaySubject then,
plus maybe add some docs for people who like to write their own StreamExtensions,
if they use transform internally, then they can be compatible, if just returning a Stream like map, then they need to add shareValue for example.

I'll try to do ReplaySubject later today

hoc081098:
@frankpepermans Should we make https://github.com/ReactiveX/rxdart/blob/5ff7fa9d372f5fae4f3e3a5f3af76eb1d51d1d9b/lib/src/utils/forwarding_stream.dart#L11 public for any custom StreamTransfomer

Issue #511, opened 2020-09-30
Can't cancel stream after switchMapping to BehaviorSubject


smkhalsa:
The following test times out:

```dart
test('can cancel stream', () async {
  final controller = StreamController();
  final stream = controller.stream
      .switchMap((value) => BehaviorSubject.seeded(true));
  controller.add(true);
  await Future.delayed(Duration.zero);
  expect(stream, emitsThrough(emitsDone));
  await controller.close();
});
```

I'd expect that closing the originating StreamController would cancel the stream. Am I misunderstanding something, or might this be a bug?

hoc081098:
because  `SwitchMap Stream` only emits done when inner `Stream` emits done 

https://github.com/ReactiveX/rxdart/blob/c546006d511bb88701de0aed35de563113df5af1/lib/src/transformers/switch_map.dart#L38

and https://github.com/ReactiveX/rxdart/blob/c546006d511bb88701de0aed35de563113df5af1/lib/src/transformers/switch_map.dart#L19-L29

smkhalsa:
Thanks @hoc081098, but what if we need to close the inner stream from outside the switchMap?

For example, in the below test `outer.close()` times out:

```dart
test('can cancel stream', () async {
  final outer = StreamController();
  BehaviorSubject inner;
  final stream = outer.stream
      .switchMap((_) => inner = BehaviorSubject.seeded(false));
  expect(stream, emitsThrough(emitsDone));
  outer.add(true);
  await Future.delayed(Duration.zero);
  await inner.close();
  await outer.close();
});
```

hoc081098:
Thanks @smkhalsa I think it is a bug, #513 will fix it

PR #513, opened 2020-10-03
fix(switchMap): doesn't close after the last inner closed

hoc081098:
:bug: Fixes https://github.com/ReactiveX/rxdart/issues/511.


frankpepermans (approved):
nice!

PR #524, opened 2020-10-29
Error and stack trace

### LINKED ISSUE #517: BehaviorSubject/ValueStream - any way to access the last error if it occurred?
`BehaviorSubject` and `ValueStream` provided a way to check whether latest value exists and get its value if it does.

Wondering is it possible to have the equivalent for the latest error?

Actually I found it shouldn't be too difficult as `_Wrapper` has already recorded the latest error and stack trace. The only thing needed is just add a few API to expose it.

hoc081098:
### ISSUE
Fixes #517 
### BREAKING CHANGE:
- Change `ValueStream.error` to `ValueStream.errorAndStackTrace`:
   using  `stream.errorAndStackTrace?.error` instead of `stream.error`
- Change `ReplayStream.errors` to `ReplayStream.errorAndStackTraces`:
   using  `stream.errorAndStackTraces.map((es) => es.error).toList()` instead of `stream.errors`
- `Notification`: merge 2 fields `error` and `stackTrace` to `errorAndStackTrace`

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -1,8 +1,9 @@
 import 'dart:async';
 
-import 'package:rxdart/rxdart.dart';
 import 'package:rxdart/src/streams/replay_stream.dart';
 import 'package:rxdart/src/streams/value_stream.dart';
+import 'package:rxdart/src/utils/error_and_stacktrace.dart';
+import 'package:rxdart/subjects.dart';
 
 /// A ConnectableStream resembles an ordinary Stream, except that it
 /// can be listened to multiple times and does not begin emitting items when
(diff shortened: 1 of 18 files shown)


lib/src/streams/connectable_stream.dart
```diff
@@ -174,7 +175,7 @@ class ValueConnectableStream<T> extends ConnectableStream<T>
   bool get hasValue => _subject.hasValue;
 
   @override
-  Object get error => _subject.error;
+  ErrorAndStackTrace get errorAndStackTrace => _subject.errorAndStackTrace;
```
narcodico, line 178: `ErrorAndStackTrace` it's a weird naming. What do you feel about `RxError` or `ErrorDetails` ?

hoc081098: 😂 It's a weird name but IMO it's clear

narcodico: It doesn't get any clearer than that 😂

frankpepermans (approved):

timnew:
The change looks great. May I ask is there any estimation that when it would go to formal version instead of beta?

PR #527, opened 2020-11-07
backpressure: max length buffer

### LINKED ISSUE #526: Debounce (and friends) should not keep the whole queue in memory until window ends
This example will pretty quickly eat all your memory:

```dart

void main(List<String> arguments) {
  final stream = Stream<void>.periodic(const Duration(microseconds: 1))
      .map((_) => 'Some data' * 1000)
      .debounceTime(Duration(seconds: 1));

  stream.listen((_) {});
}
```

hoc081098:
- Fix #526 .
- Fix travis Dart 2.12.0-13.0.dev lint: • 'onError' is deprecated and shouldn't be used. Use runZonedGuarded instead. • test/transformers/do_test.dart:302:9 • deprecated_member_use

### DIFF
--- a/lib/src/transformers/backpressure/backpressure.dart
+++ b/lib/src/transformers/backpressure/backpressure.dart
@@ -31,21 +31,24 @@ class _BackpressureStreamSink<S, T> implements ForwardingSink<S, T> {
   final bool Function(List<S> queue) _closeWindowWhen;
   final bool _ignoreEmptyWindows;
   final bool _dispatchOnClose;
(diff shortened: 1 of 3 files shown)


cbenhagen:
Wow that was fast. Thank you! I guess throttle could also use a length of 1 by default.

hoc081098:
cc @frankpepermans Would you mind reviewing?

frankpepermans (approved):
Looking good!

can't remember why we did null checks on windowStart and windowEnd?

PR #530, opened 2020-11-23
fix some API example expression error

HannibalKcc:
i found some expression error in some API's example and i try to fix them.

you can review more detail in code.


frankpepermans (approved):
thanks :)

PR #541, opened 2021-01-03
NNBD: refactor ValueStream

hoc081098:
### Assuming that:
```dart
S initial = ...;
final state$ = BehaviorSubject<S>.seed(initial);

S get state => state$.value!; // <- return type S because I always keep state$ having value. (notice: type of state$.value is S?)
```

when `S is int?`:
```dart
int? initial = null;
final state$ = BehaviorSubject<int?>.seed(initial);

int? get state => state$.value!; // <- crash, because state.value! is null.
```
will crash, but `int? get state ` should return `null`.

### This PR:
- Add `valueWrapper` to `ValueStream`.
            -  Return null if there has been no emission yet.
            -  Otherwise, return `ValueWrapper(value)` (to indicate that the latest value is null, return `ValueWrapper(null)`).
- Change `value`, `hasValue` and `hasError` to extension getters.


frankpepermans (approved):
Nice!

PR #545, opened 2021-01-27
🐛 Fix RetryWhenStream.cancel sub stream cancellation

### LINKED ISSUE #428: RetryWhenStream.cancel does not cancel the sub stream

ThomasKliszowski:
Fix #428

Previously #429, closed to re organize my fork.

### DIFF
--- a/lib/src/streams/retry_when.dart
+++ b/lib/src/streams/retry_when.dart
@@ -68,6 +68,7 @@ class RetryWhenStream<T> extends Stream<T> {
   final RetryWhenStreamFactory retryWhenFactory;
   StreamController<T> _controller;
   StreamSubscription<T> _subscription;
+  StreamSubscription<void> _subSubscription;
   final _errors = <ErrorAndStackTrace>[];
 
   /// Constructs a [Stream] that will recreate and re-listen to the source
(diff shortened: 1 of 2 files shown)


hoc081098:
Thanks @ThomasKliszowski , closes because #547 also fixed it

PR #546, opened 2021-01-27
Add TakeLastStreamTransformer and takeLast extension

### LINKED ISSUE #497: maybe we need `takeLast` ?
[rxjs takeLast](https://rxjs.dev/api/operators/takeLast).

`takeLast` is a useful `API`，but i didn't find a similar `API` in `rxdart` or `dart stream`.can we provide this `API` in future?

TKS for any help.

ThomasKliszowski:
Fix #497


ThomasKliszowski:
Codecov does not update, the coverage is 100%, and the global coverage diff is +0.06%.

https://codecov.io/gh/ReactiveX/rxdart/commit/a210c26a65ae4d09bc4a4373fb2db63a8bcf2b68/

lib/src/transformers/take_last.dart
```diff
@@ -0,0 +1,81 @@
...
+
+  @override
+  void close(EventSink<T> sink) {
+    buffer.forEach(sink.add);
+    buffer = [];
+    sink.close();
+  }
+
+  @override
+  FutureOr onCancel(EventSink<T> sink) {}
```
hoc081098, line 34: Should clear list/queue on cancel.

lib/src/transformers/take_last.dart
```diff
@@ -0,0 +1,81 @@
...
+
+/// Extends the [Stream] class with the ability receive only the final [count]
+/// events from the source [Stream].
+extension TakeLastExtension<T> on Stream<T> {
+  /// Emits only the final [count] values emitted by the source [Stream].
+  ///
+  /// ### Example
+  ///
+  ///     Stream.fromIterable([1, 2, 3, 4, 5])
+  ///       .transform(TakeLastStreamTransformer(3))
```
hoc081098, line 77: Should be `Stream.fromIterable([1, 2, 3, 4, 5]).takeLast(3)`. Because it is example for `takeLast` extension method :))

test/transformers/take_last_test.dart
```diff
@@ -0,0 +1,92 @@
...
+        .listen(expectAsync1((data) {
+      expect(data, 3);
+      subscription.cancel();
+    }));
+
+    subscription.pause();
+    subscription.resume();
+  });
+
+  test('Rx.takeLast.accidentalBroadcast', () async {
```
hoc081098, line 82: IMO, `Rx.takeLast.singleSubscription` or `Rx.takeLast.single.subscription` is better name.

hoc081098 (changes requested):

lib/src/transformers/take_last.dart
```diff
@@ -0,0 +1,81 @@
+import 'dart:async';
+
+import 'package:rxdart/src/utils/forwarding_sink.dart';
+import 'package:rxdart/src/utils/forwarding_stream.dart';
+
+class _TakeLastStreamSink<T> implements ForwardingSink<T, T> {
+  _TakeLastStreamSink(this.count);
+
+  final int count;
+  List<T> buffer = [];
```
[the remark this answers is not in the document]
ThomasKliszowski: Way better indeed.

lib/src/transformers/take_last.dart
ThomasKliszowski: `onCancel` can be called multiple times, right? Could be a problem if the queue is cleared?

lib/src/transformers/take_last.dart
ThomasKliszowski: Good catch.

lib/src/transformers/take_last.dart
hoc081098: When cancelling subscription or calling sink.close(), onCancel will be called.
For broadcast stream, onCancel can be called multiples times, but we only focus from `onListen` untill `onCancel`.

For example, I copied from `sdk`:
```dart
   * The returned stream is a broadcast stream if this stream is.
   * For a broadcast stream, the events are only counted from the time
   * the returned stream is listened to.
   */
  Stream<T> skip(int count) {
    return new _SkipStream<T>(this, count);
  }
```

```suggestion

  @override
  void close(EventSink<T> sink) {
    queue.forEach(sink.add);
    sink.close(); // <- also call onCancel
  }

  @override
  FutureOr onCancel(EventSink<T> sink) => queue.clear();

```

ThomasKliszowski: OK that's perfect, thanks!

hoc081098 (approved):

PR #547, opened 2021-01-30
Reworks retry/retryWhen

### LINKED ISSUE #544: RetryWhen's RetryError is not a transparent solution
So I have a peeve with RetryError. My problem with it is that it wraps and thus obfuscates the original error. This in turn makes retryWhen non-transparent: you can't insert retryWhen into an existing stream hierarchy and expect it to work as it used to work, because exceptions now get transformed to RetryError.

I ran into a scenario like the above, where I had to inject retry logic into existing stream(s), but ended up having to refactor out the RetryError from the existing RxDart code, to make it behave as expected.

Can we discuss if it would be possible to remove RetryError from this operator? It does not belong IMO, and unhandled errors should never be scooped up mid-transit: they are important for breaking out of the control flow.

hoc081098:
Fixes #544, #428

- Remove `RetryError` (Because Dart Stream can deliver more than one error).
- Changes for `retry`: add all errors if retry fails. 
- Changes for `retryWhen`: add original error, and error from factory if they are not identical. 
- Change `StackTrace?` to `StackTrace` because it is not null, 
https://github.com/dart-lang/sdk/blob/53f5179b45911fd1f937239f91e55ec8badd7975/sdk/lib/async/stream.dart#L461-L462
```dart
  /// The [onError] callback must be of type `void Function(Object error)` or
  /// `void Function(Object error, StackTrace)`.
  /// The function type determines whether [onError] is invoked with a stack
  /// trace argument.
  /// The stack trace argument may be [StackTrace.empty] if this stream received
  /// an error without a stack trace.
```


frankpepermans (approved):
looks nice!

PR #552, opened 2021-02-23
prepare null-safe, stable release

kevmoo:


frankpepermans:
Thanks!
@hoc081098 let's include this and your PR's, then do a next release?

Issue #550, opened 2021-02-08
Non-nullable-by-default value streams


ryanheise:
A seeded `ValueStream` (e.g. as produced via `BehaviorSubject.seeded`) can in principle always have a non-null value, but this is unfortunately not reflected by the type system and consequently requires null checks on `stream.value`. In an app where all value streams are seeded, it can be quite inconvenient that the the nullsafe features of the language don't help.

Because `T` is a subtype of `T?` for all `T`, we can solve this problem by having a seeded subclass of `ValueStream`:

```dart
abstract class SeededValueStream<T> extends ValueStream<T> {
   @override
   T get value => super.value!;
}
```

(I understand rxdart now implements `value` as an extension method, but this is just to illustrate the general principle of how the typing could work.)

We can then have the `seeded` constructors return `SeededValueStream`s.

hoc081098:
But when callling `BehaviorSubject.addError(e, s)`, `value` returns `null` 😃

ryanheise:
Then, the `SeededValueStream.value` getter could be made to throw the actual stream error. But that would violate the substitution principle if `SeededValueStream` is a subclass of `ValueStream`, or at least would be a breaking API change.

If you take a look at the `AsyncSnapshot` class in the standard libraries, it implements both approaches. It provides a getter that returns `T?`and then provides another getter that returns `T` which either returns the value or throws the error if there is an error.

hoc081098:
Why not use `ValueStream.requireValue`:
https://github.com/ReactiveX/rxdart/blob/7a1c89b386a90d8d19ee8cfff3feaf85ef9e0085/lib/src/streams/value_stream.dart#L27

ryanheise:
Ah yes, that is exactly what I was looking for :-). Apologies, I am still navigating my way around the new code and extension methods and didn't notice it already does this. I'll close this issue.

ryanheise:
Although I've closed the issue, I would say I do slightly prefer the `AsyncSnapshot` API where `requireData` throws the actual underlying stream error, and not merely a null check assertion.

hoc081098:
See PR #551

PR #560, opened 2021-03-09
Refactor ValueStream: remove ValueWrapper

### LINKED ISSUE #556: Whats the motivation for creating ValueWrapper and breaking backwards compatibility with ValueStream?
`ValueWrapper` seems to do nothing other than wrap another value, which `ValueStream` also did.

So why create it? Was it really necessary to break everyone's code due to this seemingly useless class? I'm genuinely curious as to why this was added.

So our code is going from `_valueStream.value` to `_valueStream.valueWrapper.value`.

hoc081098:
Fixes #556 
- #559 ValueStream
  - `ValueWrapper` -> `[requireValue, valueOrNull, hasValue]`.
  - `errorAndStackTraces` -> [`requireError`, `errorOrNull`, `stackTraceOrNull`].
- ReplayStream: `errorAndStackTraces` -> [`errors`, `stackTraces`].
- Keep `ValueWrapper` as internal implementation (to able to keep `null` as the latest value when `T` is nullable type).

### DIFF
--- a/lib/src/streams/value_stream.dart
+++ b/lib/src/streams/value_stream.dart
@@ -1,12 +1,9 @@
-import 'package:rxdart/src/utils/error_and_stacktrace.dart';
-
 /// An [Stream] that provides synchronous access to the last emitted item
 abstract class ValueStream<T> implements Stream<T> {
   /// Returns the last emitted value, failing if there is no value.
   /// See [hasValue] to determine whether [value] has already been set.
   ///
-  /// Throws [error] if [hasError].
-  /// Throws [StateError], if neither [hasValue] nor [hasError].
+  /// Throws [ValueIsMissingError] if has no value or has error.
   T get value;
 
   /// Returns either [value], or `null`, should [value] not yet have been set.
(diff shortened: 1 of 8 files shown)


lib/src/streams/value_stream.dart
```diff
@@ -16,38 +23,15 @@ abstract class ValueStream<T> implements Stream<T> {
...
-    throw StateError('Neither data event nor error event has been emitted.');
-  }
-
   /// A flag that turns true as soon as at an error event has been emitted.
   bool get hasError => errorAndStackTrace != null;
 
-  /// Last emitted error, or null if no error added or value exists.
-  Object? get error => errorAndStackTrace?.error;
+  /// Last emitted error, or `null` if no error added or value exists.
+  Object? get errorOrNull => errorAndStackTrace?.error;
```
frankpepermans, line 30: Perhaps we should move error and errorOrNull also outside of an extension method, and just inside ValueStream?

lib/src/subjects/behavior_subject.dart
```diff
@@ -136,7 +136,28 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
   ValueStream<T> get stream => this;
 
   @override
-  ValueWrapper<T>? get valueWrapper => _wrapper.latestValue;
+  bool get hasValue => _wrapper.latestValue != null;
+
+  @override
+  T get value {
```
frankpepermans, line 142: I'd just return `_wrapper.latestValue!` here, if T is non-nullable, then a developer expects that getting value would throw in this way if it is null, a `StateError` is confusing.

Also, I don't think we should check for any existing errors either, as just value is being accessed.

lib/src/subjects/behavior_subject.dart
```diff
@@ -136,7 +136,28 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
...
+    }
+
+    throw StateError('Neither data event nor error event has been emitted.');
+  }
+
+  @override
+  T? get valueOrNull => _wrapper.latestValue?.value;
+
+  /// Set and emit the new value.
+  set value(T newValue) => add(newValue);
```
frankpepermans, line 160: Do we want a setter for value?

imagine:
```dart
subject.value = 1;
expect(subject.value, 1); // can throw, because the internal add is async
```

frankpepermans: IMO, `value` is actually dangerous, it's forcing sync access to data inside a `Stream`, but we cannot guarantee its behavior, using `value` is really as a reliable source is something I think we should actively discourage.

lib/src/subjects/behavior_subject.dart
hoc081098: Currently, `BehaviorSubject.hasValue` will be false if the last event is an error. Should we separator data from error?

lib/src/subjects/behavior_subject.dart
hoc081098: > Do we want a setter for value?
> 
> imagine:
> 
> ```dart
> subject.value = 1;
> expect(subject.value, 1); // can throw, because the internal add is async
> ```

This always works, because value is set synchronouslly

lib/src/subjects/behavior_subject.dart
hoc081098: @frankpepermans look at https://github.com/ReactiveX/rxdart/issues/550#issuecomment-775193768 , should we throw "a custom error" (eg. `ValueStreamError`) to indicate that the Stream has no value.

lib/src/subjects/behavior_subject.dart
frankpepermans: True, then ignore my remark

lib/src/streams/value_stream.dart
hoc081098: Yes, should move all of them into `ValueStream` class

lib/src/subjects/behavior_subject.dart
frankpepermans: Yes, I think so, if you do stream.last for example, but no values are emitted, it also throws.

Stream.last and value are not 100% the same of course, but then it would be consistent

frankpepermans: As for the error, hasValue maybe needs to look at value only, so if value is filled and an error occurs, don't change hasValue.

You may want to do something with value on error

frankpepermans (approved):

PR #565, opened 2021-03-21
`Connectable`: remove unnecessary `asBroadcastStream`

hoc081098:
`source.asBroadcastStream(onCancel: (s) => s.cancel())` is same as `share()`, convert to broadcast stream twice is unnecessary.


hoc081098:
Seem be a mistake 😆

PR #567, opened 2021-03-26
`Connectable`: remove unnecessary `asBroadcastStream`

hoc081098:
Just cache `ConnectableStreamSubscription`, `source` will only be listened once

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -51,22 +51,21 @@ class PublishConnectableStream<T> extends ConnectableStream<T> {
 
   PublishConnectableStream._(this._source, this._subject) : super(_subject);
 
-  ConnectableStreamSubscription<T> _connect() =>
-      ConnectableStreamSubscription<T>(
-        _source.listen(
-          _subject.add,
-          onError: _subject.addError,
-          onDone: _subject.close,
-        ),
-        _subject,
-      );
+  late final _connection = ConnectableStreamSubscription<T>(
+    _source.listen(
+      _subject.add,
+      onError: _subject.addError,
+      onDone: _subject.close,
+    ),
+    _subject,
+  );
 
   @override
   Stream<T> autoConnect({
     void Function(StreamSubscription<T> subscription)? connection,
   }) {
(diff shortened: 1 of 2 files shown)


lib/src/streams/connectable_stream.dart
```diff
@@ -128,22 +119,21 @@ class ValueConnectableStream<T> extends ConnectableStream<T>
...
-  ConnectableStreamSubscription<T> _connect() =>
-      ConnectableStreamSubscription<T>(
-        _source.listen(
-          _subject.add,
-          onError: _subject.addError,
-          onDone: _subject.close,
-        ),
-        _subject,
-      );
+  late final _connection = ConnectableStreamSubscription<T>(
```
frankpepermans, line 122: Why is the late required here?

lib/src/streams/connectable_stream.dart
```diff
@@ -49,28 +49,23 @@ class PublishConnectableStream<T> extends ConnectableStream<T> {
...
-          _subject.add,
-          onError: _subject.addError,
-          onDone: _subject.close,
-        ),
-        _subject,
-      );
+  PublishConnectableStream._(this._source, this._subject) : super(_subject);
+
+  late final _connection = ConnectableStreamSubscription<T>(
+    _source.listen(
```
frankpepermans, line 55: Also unnecessary late

lib/src/streams/connectable_stream.dart
hoc081098: Make it gets computed only upon first access. We should not listen to the source before `onListen` or `connect`  :)) https://dart.dev/null-safety/understanding-null-safety#late-final-variables

frankpepermans: But you're immediately assigning a value here, afaik, properties are late by default, you could drop late here

hoc081098: Cannot drop :)) 
image

frankpepermans: Urgh ok then :)

frankpepermans (approved):

PR #570, opened 2021-03-27
`ValueStream` contains only value, error and stackTrace

hoc081098:
#556 
```dart
/// An [Stream] that provides synchronous access to the last emitted item
abstract class ValueStream<T> implements Stream<T> {
  /// Returns the last emitted value, failing if there is no value.
  /// See [hasValue] to determine whether [value] has already been set.
  ///
  /// Throws [ValueStreamError] if this Stream has no value.
  T get value;

  /// Returns last emitted error, failing if there is no error.
  ///
  /// Throws [ValueStreamError] if this Stream has no error.
  Object get error;

  /// Returns [StackTrace] of the last emitted error,
  /// or `null` if no error added or the added error has no [StackTrace].
  StackTrace? get stackTrace;
}
```

### DIFF
--- a/lib/src/streams/connectable_stream.dart
+++ b/lib/src/streams/connectable_stream.dart
@@ -158,22 +158,10 @@ class ValueConnectableStream<T> extends ConnectableStream<T>
   }
 
   @override
-  bool get hasValue => _subject.hasValue;
-
-  @override
   T get value => _subject.value;
 
   @override
-  T? get valueOrNull => _subject.valueOrNull;
-
-  @override
   Object get error => _subject.error;
-
-  @override
-  Object? get errorOrNull => _subject.errorOrNull;
-
-  @override
-  bool get hasError => _subject.hasError;
 
   @override
   StackTrace? get stackTrace => _subject.stackTrace;
--- a/lib/src/streams/value_stream.dart
+++ b/lib/src/streams/value_stream.dart
@@ -6,26 +6,55 @@ abstract class ValueStream<T> implements Stream<T> {
   /// Throws [ValueStreamError] if this Stream has no value.
   T get value;
 
-  /// Returns either [value], or `null`, should [value] not yet have been set.
-  T? get valueOrNull;
-
-  /// Returns `true` when [value] is available.
-  bool get hasValue;
-
   /// Returns last emitted error, failing if there is no error.
   ///
   /// Throws [ValueStreamError] if this Stream has no error.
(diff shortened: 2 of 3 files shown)


larssn:
```
  @override
  T get value {
    final wrapper = _wrapper.value;
    if (wrapper != null) {
      return wrapper.value;
    }
    throw ValueStreamError.hasNoValue();
  }
```

This just looks like indirection and indirection is not abstraction. I can see the idea behind it. But it feels like we're just bending over backwards again to try and get around that null check, by adding additional fields and wrappers. Does that solve anything? I would say no, as the user experience is the same. In both cases you either have to check for a null or now - an error.

Now `ValueStreamExtensions`. This was created to satisfy the people who needs those getters? In that case I'd just put them on the main class. My personal opinion is you rarely need extension methods internally in library, as you have access to all the code anyway.

@frankpepermans I think forcing `value` to be non-nullable should be reconsidered. It's causing a lot of unnecessary backflips through loops and hoops.

The fact is that if unseeded, `value` **will** be null, so we're coding around a valid crash. Sure we could just try and catch that error, but we'll just end up throwing another one. So we're just substituting one error for another. Correct me if I'm wrong on that.

What should happen here:
```
test('value test', () async {
      final unseeded = BehaviorSubject<int>();
      // Which of the following tests are we aiming for? And what does throwing a ValueStreamError error give us? It does not offload the consumer.
      expect(unseeded.value, throwsA(ValueStreamError));
      expect(unseeded.value, null);
      // expect(unseeded.value, 'what happens???');
    });
```

The null test is the idiomatic approach IMO.

jyardin:
About the implementation : I would not have moved `hasValue` in the extension. You have to rely on throwing and catching an exception for a simple check, which is not cheap in terms of performance.

xvrh:
As a user of this library, I feel like an extension is not the right choices for those getters. It forces users of this library to re-export `rxdart` each time we want to expose a `ValueStream` in a public API. Which is really annoying.

jyardin:
To add to my previous comment: it might make sense to move `valueOrNull` in an extension, but not `hasValue` in my opinion. 
`hasValue` is essential to the contract of the class as you cannot safely call `value` without checking `hasValue` first.

That said, I still agree with @larssn that a nullable value (so basically renaming `valueOrNull` to `value`) would be a simpler contract (without runtime exception).

hoc081098:
So basically, keep the old implementation 😃

Issue #587, opened 2021-05-24
Deadlocks in tests after upgrading rxdart


Mike278:
Took another stab at upgrading off rxdart 0.23.1, but unfortunately still getting some failing tests. Here's a repro/comparison (could probably be minimized further but not sure which parts are relevant).

```dart
import 'package:meta/meta.dart';
import 'package:moor/ffi.dart';
import 'package:moor/moor.dart';
import 'package:rxdart/rxdart.dart';
import 'package:test/test.dart';

class TestDb extends GeneratedDatabase {
  TestDb() : super(SqlTypeSystem.withDefaults(), VmDatabase.memory());
  @override final List<TableInfo> allTables = const [];
  @override final int schemaVersion = 1;
}

Future<void> testCase({
  @required Stream<int> Function() createOuter,
  @required Stream<int> Function() createInner,
}) async {
  final log = <int>[];
  final timeout = Duration(milliseconds: 100);

  final a = createOuter();
  final b = a.switchMap((_) => createInner());

  b.listen(log.add);
  await b.first.then(log.add)
    .timeout(timeout, onTimeout: () => fail('1st should complete'));
  expect(log, [2, 2]);

  b.listen(log.add);
  await b.first.then(log.add)
    .timeout(timeout, onTimeout: () => fail('2nd should complete'));
  expect(log, [2, 2, 2, 2]);
}

void main() {
  group('rxdart upgrade', () {
  
    test("moor", () async {
      final db = TestDb();
      Stream<int> selectInt(int i) => db
        .customSelect('select $i a')
        .map((row) => row.read<int>('a'))
        .watchSingle();
      await testCase(
        createOuter: () => selectInt(1),
        createInner: () => selectInt(2),
      );
    });

    test("rxdart", () async {
      final outer = BehaviorSubject<int>();
      final tc = testCase(
        createOuter: () => outer,
        createInner: () {
          final inner = BehaviorSubject<int>();
          Future.delayed(Duration(milliseconds: 10)).then((_) => inner.add(2));
          return inner;
        }
      );
      await Future.delayed(Duration(milliseconds: 10));
      outer.add(1);
      await tc;
    });
  });
}
```

- rxdart 0.23.1: Both tests pass
- rxdart 0.24.0 and 0.24.1: Both tests fail with "2nd should complete"
- rxdart 0.25.0, 0.26.0, and 0.27.0: Only the moor test fails
[...]

PR #588, opened 2021-05-24
Change behavior in forward stream

frankpepermans:
As reported here: https://github.com/ReactiveX/rxdart/issues/587

We currently have an issue that any transformer which uses `forwardStream`, only invokes `listen` once, and afterwards delegates `listen` calls to its inner `controller.stream`.

Best described with a test:
```dart
void main() {
  final controller = StreamController<void>.broadcast();
  final stream = DoOnSubscribeStream(controller.stream, () {
    print('new subscription');
  });

  var switched = stream.switchMap((event) => Stream.value(event));
  switched.listen(null); // prints!
  switched.listen(null); // does not print!

  var mapped = stream.map((event) => event);
  mapped.listen(null); // prints!
  mapped.listen(null); // prints!
}
```
If we run this, then the first use case with `switchMap` only triggers the listen _once_, whereas the `map` version triggers is _twice_

If a user overrides the `listen` method, then this becomes problematic.

### DIFF
--- a/lib/src/utils/forwarding_stream.dart
+++ b/lib/src/utils/forwarding_stream.dart
@@ -12,7 +12,8 @@ class ForwardingStream<T, R> extends Stream<R> {
   final Stream<T> _inner;
   final ForwardingSink<T, R> _connectedSink;
   late StreamController<R> _controller;
-  StreamSubscription<T>? _subscription;
+  late StreamSubscription<T> _subscription;
+  bool _didListenOnce = false;
 
   /// Constructs a ForwardingStream using an underlying "inner" [Stream],
   /// and a [ForwardingSink], which will connect to events.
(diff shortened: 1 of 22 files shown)


lib/src/utils/forwarding_stream.dart
```diff
@@ -81,5 +71,12 @@ Stream<R> forwardStream<T, R>(
     );
   }
 
+  subscription = stream.listen(
```
hoc081098, line 74: I think this is wrong. We must listen to upstream when downstream is listen to.

frankpepermans: True, this is too eager,

I'll wrap it in a ForwardingStream instead, and then override its listen method, then return that Stream instead.

frankpepermans: Urgh this needs more time, I'll revisit asap

hoc081098:
@frankpepermans I don't see any difference between `ForwardingStream` class vs `forwardStream` method :))

lib/src/utils/forwarding_stream.dart
hoc081098: Current implementation is them same as `transformByHandlers ` from `stream_transform` https://github.com/dart-lang/stream_transform/blob/533e1af154629959545e3a3728312af6cb4f5619/lib/src/from_handlers.dart#L13

hoc081098: built-in operator in Stream class does not cache subscription and state for Broadcast Stream, but `rxdart` and `stream_transform` do it, I think problem caused by `map` operator

frankpepermans:
@hoc081098 idd, we still only subscribe to the inner Stream once,
we kinda want to do that each time, but still yield the outer (controller.stream) subscription instead.

Otherwise, things like the doOnX would trigger too often for example.

So I made an ugly, ugly hack for now...

Subscribe on listen and forward to the controller,
when another listen enters, listen again, but use a silent one, and return a 'combined' subscription.

...I just pushed this to kinda see where we need to go from here

PR #594, opened 2021-05-30
Added `Single` stream, it always either emits one value or an error notification.

BreX900:
Added `Single` is something like an `Stream`, but instead of emitting a series of values it always either emits one value or an error notification.

This class is similar to: http://reactivex.io/documentation/single.html


hoc081098:
I see this is quite similar to https://github.com/hoc081098/rxdart_ext/tree/master/lib/src/single. Should we focus on `rxdart_ext` and some time in the future, merge back to `rxdart`?

PR #601, opened 2021-06-14
Use Stream.multi instead of forwardStream's inner Controller

frankpepermans:
Fixes the issue where we rely on specific `createForwardingStream`'s, thus respecting the behavior of custom `StreamController`s, see: https://github.com/ReactiveX/rxdart/issues/587

### DIFF
--- a/lib/src/utils/forwarding_stream.dart
+++ b/lib/src/utils/forwarding_stream.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:math';
 
 import 'package:rxdart/src/utils/forwarding_sink.dart';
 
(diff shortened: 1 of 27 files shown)


frankpepermans:
_Some_ tests have been skipped, they all use Timers and were flaky by design.
I'm gonna look for a better way to test them.

lib/src/utils/forwarding_stream.dart
```diff
@@ -1,85 +1,146 @@
...
+          _runCatching(() => _connectedSink.onPause(_compositeController));
+        };
+        controller.onResume = () {
+          _subscription?.resume();
+          _runCatching(() => _connectedSink.onResume(_compositeController));
+        };
+        controller.onCancel = () {
+          final onCancelConnectedFuture =
+              _connectedSink.onCancel(_compositeController);
+          final onCancelSubscriptionFuture = _subscription?.cancel();
```
hoc081098, line 100: Should we cancel subscription when canceling MultiStreamController?

PR #605, opened 2021-06-19
Refactor forwardStream: uses `Stream.multi`

### LINKED ISSUE #587: Deadlocks in tests after upgrading rxdart

hoc081098:
Based on #596 and #601 

- Remove try catch to increase performance.
- Implementation classes must handle error.
- Uses Stream.multi (Fixes #587)
- doX operator uses normal broadcast StreamController


hoc081098:
A special case cannot be handled :((
```dart
    test('description', () async{
      final a = BehaviorSubject.seeded('b');
      final bug = a.doOnData((v) {});
      await bug.listen(null).cancel();
      expect(await bug.first, 'b'); // cannot completes
    });
```

frankpepermans:
I also tried converting to Stream.multi, but where my attempt fails, so does this one, take this test for example:

```dart
test('issue/587', () async {
        final source = BehaviorSubject.seeded('source');
        final switched =
            source.switchMap((value) => BehaviorSubject.seeded('switched'));
        var i = 0;
        switched.listen((_) => i++);
        expect(await switched.first, 'switched');
        expect(i, 1);
        expect(await switched.first, 'switched');
        expect(i, 1);
      });
```

hoc081098:
> I also tried converting to Stream.multi, but where my attempt fails, so does this one, take this test for example:
> 
> ```dart
> test('issue/587', () async {
>         final source = BehaviorSubject.seeded('source');
>         final switched =
>             source.switchMap((value) => BehaviorSubject.seeded('switched'));
>         var i = 0;
>         switched.listen((_) => i++);
>         expect(await switched.first, 'switched');
>         expect(i, 1);
>         expect(await switched.first, 'switched');
>         expect(i, 1);
>       });
> ```

I have just skipped some tests:
- 'issue/419: sync throughput'
- 'Rx.sample.reusable'
- 'Rx.startWith broadcast stream should not startWith on multiple subscribers'

frankpepermans (approved):
Sorry for the delay again!

Awesome job yet again :)

And great to hear the deadlocks are solved!

PR #627, opened 2021-09-13
fix(example): SearchResultItem.fromJson type error

wayne900204:
If it uses Object will error.
My sdk is not >2.14, So I am not sure it is a bug, but I think it is a bug. because I searched history commit, no one changed this.  see: #628


hoc081098 (approved):
This error has occurred since null-safety is enabled.
Thanks @wayne900204

PR #633, opened 2021-09-24
Improved CompositeSubscription

BreX900:
- CompositeSubscription's `dispose`, `clear`, and `remove` methods now return a completion future
- Fixed an issue where a stream not present in CompositeSubscription was canceled
- Added the ability not to cancel the stream when it is removed from CompositeSubscription

### DIFF
--- a/lib/src/utils/composite_subscription.dart
+++ b/lib/src/utils/composite_subscription.dart
@@ -48,26 +48,32 @@ class CompositeSubscription {
     return subscription;
   }
 
-  /// Cancels subscription and removes it from this composite.
-  void remove(StreamSubscription<dynamic> subscription) {
-    subscription.cancel();
-    _subscriptionsList.remove(subscription);
+  /// Remove the subscription from this composite and cancel it if it has been removed.
+  Future<void>? remove(
+    StreamSubscription<dynamic> subscription, {
+    bool shouldCancel = true,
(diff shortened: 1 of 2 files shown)


lib/src/utils/composite_subscription.dart
```diff
@@ -48,26 +48,32 @@ class CompositeSubscription {
...
   }
 
-  /// Cancels subscription and removes it from this composite.
-  void remove(StreamSubscription<dynamic> subscription) {
-    subscription.cancel();
-    _subscriptionsList.remove(subscription);
+  /// Remove the subscription from this composite and cancel it if it has been removed.
+  Future<void>? remove(
+    StreamSubscription<dynamic> subscription, {
+    bool canCancel = true,
```
hoc081098, line 54: I think we should use `bool shouldCancel = true` :))

BreX900:
@hoc081098 Wouldn't it be more correct to use `FutureOr<void>` instead of `Future<void>?` ?

hoc081098:
> @hoc081098 Wouldn't it be more correct to use `FutureOr<void>` instead of `Future<void>?` ?

I think 'Future<void>?' is good, since null-safety is enabled everywhere 😂

hoc081098 (approved):
Thanks for your contribution!

PR #637, opened 2021-09-29
README fix

purplenoodlesoop:
Fix of the single line in the README.md. 

The `Stream<T>.value(...)` constructor was added in the 2.5.0 version of the SDK, and that part of the README was a bit misleading.


hoc081098:
I think that is about TYPE and BEHAVIOR, only not behavior

hoc081098:
Thanks, but closing via #638

PR #666, opened 2022-03-11
Add BufferedSubject

tjarvstrand:


tjarvstrand:
Is anyone even looking at these PRs? 

@hoc081098 You seem to be the most active committer in this repository lately. Do you have any opinions?

hoc081098:
Hi @tjarvstrand, I think it is not a standard Subject in ReactiveX

tjarvstrand:
Ah, is that a requirement?

hoc081098:
Yes.

- A similar case:
This subject is similar to `DispatchWorkSubject`, but `DispatchWorkSubject ` is in [RxJavaExtensions](https://github.com/akarnokd/RxJavaExtensions/blob/3.x/src/main/java/hu/akarnokd/rxjava3/subjects/DispatchWorkSubject.java) instead of RxJava
- You can include this subject in a other library with a easy effort. For me, I have created [StateSubject](https://pub.dev/documentation/rxdart_ext/latest/state_stream/StateSubject-class.html) and included it in my library [rxdart_ext](https://pub.dev/packages/rxdart_ext)
- We should keep `rxdart` as minimal as possible.

PR #674, opened 2022-05-15
ci: remove deprecated `packages` flag

beeth0ven:
Hi there.

Form `coverage`  [v1.3.0 - 2022-5-11 changelog](https://github.com/dart-lang/coverage/blob/master/CHANGELOG.md#130---2022-5-11):
>Add a --package flag, which takes the package's root directory, instead of the .package file. Deprecate the --packages flag.

Our ci raised an issue the same as dart-lang/coverage#382, the quick fix would be remove deprecated `packages` flag as metioned [here](https://github.com/dart-lang/coverage/issues/382#issuecomment-1125493001).

Thanks.

### DIFF
--- a/.github/workflows/rxdart-test.yml
+++ b/.github/workflows/rxdart-test.yml
@@ -66,7 +66,6 @@ jobs:
           --lcov
           --in=coverage.json
           --out=lcov.info
-          --packages=.packages
           --report-on=lib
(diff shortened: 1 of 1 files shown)


hoc081098 (approved):
Thanks @beeth0ven

PR #682, opened 2022-06-14
Fix ConcatEagerStream bug

Stitch-Taotao:
activeSubscription never change to next subsciption.


hoc081098 (approved):
Thanks @Stitch-Taotao

PR #687, opened 2022-07-05
add isClose

SpiciedCrab:
Add isClose verification before _add, avoiding setting newValue when the subject is closed.


hoc081098:
thank @SpiciedCrab, PR #698 continues this PR

PR #699, opened 2022-11-16
refactor(subject): `Subject.stream` now returns a read-only `Stream`

hoc081098:
- Streams returned by Subjects are read-only streams, ie. they don't support adding events (Previous, `Subject.stream` is identical to Subject).
 - Change return type of `ReplaySubject<T>.stream` to `ReplayStream<T>`

### DIFF
--- a/lib/src/subjects/behavior_subject.dart
+++ b/lib/src/subjects/behavior_subject.dart
@@ -132,7 +132,7 @@ class BehaviorSubject<T> extends Subject<T> implements ValueStream<T> {
       _wrapper.setError(error, stackTrace);
 
   @override
-  ValueStream<T> get stream => this;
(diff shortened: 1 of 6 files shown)


ryanheise:
I would prefer to leave this sort of thing the responsibility of static type checking. Adding defensive programming here increases the binary size and slightly adds to the computation time, whereas static type checking serves the same purpose but costs nothing at runtime.

What I mean regarding the static typing is that the return type of `Subject.stream` is already `ValueStream<T>` and so it already restricts the visible interface to the intended one. If a client does try to override the static types at runtime by casting that return type to some undocumented implementation type, it is using undocumented behaviour in which case the library is no longer responsible for perils the client suffers. There is no need to add executable code to defend against clients that try to bypass the declared static types.

PR #705, opened 2022-11-28
Adding FlatMapLast

pomarec:


hoc081098:
I see it is the same as `switchMap` 🤔

pomarec:
Yes i though so aswell but switch map keeps emiting from previous generated streams.
In the first test it would emit ['a-1', 'b-1', 'a-2', 'b-2', emitsDone]

hoc081098:
@pomarec, switchMap cancels previous inner stream and start listening to new inner stream.

```dart
  await Stream.fromFutures([
      Future.delayed(Duration(milliseconds: 1000), () => 'a'),
      Future.delayed(Duration(milliseconds: 2000), () => 'b')
    ]).switchMap<void>((String value) => Stream.fromFutures([
        Future.delayed(Duration(milliseconds: 300), () => '$value 1 '),
        Future.delayed(Duration(milliseconds: 1500), () => '$value 2 '),
    ])).forEach(print);  // prints a1, b1, b2
```

pomarec:
My bad, we got lost with my workmate. Thanks for your attention.

Issue #709, opened 2023-01-13
throttled stream ignores the close command


aurokk:
Hey! 
I was playing around with rxdart and faced a problem. 
When throttleTime is used then the stream ignores *the close command*. 
Is that by design or I'm doing sth wrong?

I use the latest versions of rxdart (0.27.7)/flutter (3.3.10) :-)

// WORKS
```dart
main() async {
  final c = StreamController<String>();
  Future.delayed(Duration(seconds: 5)).then((f) => c.close());
  final s = c.stream;
  print("start");
  await for (var _ in s) {
    print("loop");
  }
  print("end");
}
```

// WORKS
```dart
main() async {
  final c = StreamController<String>();
  Future.delayed(Duration(seconds: 5)).then((f) => c.close());
  final s = c.stream.debounceTime(Duration(seconds: 1));
  print("start");
  await for (var _ in s) {
    print("loop");
  }
  print("end");
}
```

// DOESN'T WORK, NEVER REACHES THE END
```dart
main() async {
  final c = StreamController<String>();
  Future.delayed(Duration(seconds: 5)).then((f) => c.close());
  final s = c.stream.throttleTime(Duration(seconds: 1));
  print("start");
  await for (var _ in s) {
    print("loop");
  }
  print("end");
}
```

PR #710, opened 2023-01-20
fix(throttle): fix issue #709, where throttled Stream does not emit done event.

hoc081098:
…one event.


hoc081098:
@frankpepermans I'm not sure this change is correct. Could you review it, please? Thanks 👍

hoc081098:
@copilot resolve conflict and continue the previous work

PR #712, opened 2023-02-22
fix example on README

wurikiji:
The default behavior of the `throttle` and `throttleTime` is to emit the leading item of the time window.  The implementation is correct, but the example of the README needs to be corrected. The example should print `1`, not `3`.


hoc081098 (approved):
Good catching

PR #728, opened 2023-06-12
add ability to identify if latest data in the stream was value

### LINKED ISSUE #603: BehaviorSubject not cleaning error after added data
I don't know why but when I add an error to BehaviorSubject and after add some data. It not cleaning any more the error. And I can't pass null to addError,  So BehaviorSubject.hasError is always true.

TatsuUkraine:
Provides proxy getter for Wrapper.isValue field, to identify if most recent data in the stream was value

fixes #603


TatsuUkraine:
Closing due to #729

PR #744, opened 2024-01-09
When onErrorResumeNext is used, error do not cancel to listen previous stream

rouddy:
When onErrorResumeNext is used, error do not cancel to listen previous stream.

here's test code

```dart
Rx.merge([
      Stream.periodic(const Duration(seconds: 1), (i) => i).map((event) => event + 1),
      Rx.concat([
        Rx.timer(1, const Duration(seconds: 5)),
        Stream.error(Exception("test"))
      ]).ignoreElements(),
    ])
        .asBroadcastStream()
        .doOnError((p0, p1) => print("!!!!! doOnError:$p0"))
        .onErrorResumeNext(Stream.periodic(const Duration(seconds: 1), (i) => i * 10))
        .listen((event) {
      print("!!!!! $event");
    }, onDone: () {
      print("!!!!! onDone");
    }, onError: (error, stacktrace) {
      print("!!!!! onError $error, $stacktrace");
    });
```

thanx

### DIFF
--- a/lib/src/utils/forwarding_stream.dart
+++ b/lib/src/utils/forwarding_stream.dart
@@ -37,6 +37,7 @@ Stream<R> _forwardMulti<T, R>(
         sink.onData,
         onError: sink.onError,
         onDone: sink.onDone,
+        cancelOnError: true,
       );
       sink.setSubscription(subscription);
     }
@@ -86,6 +87,7 @@ Stream<R> _forward<T, R>(
         sink.onData,
         onError: sink.onError,
         onDone: sink.onDone,
+        cancelOnError: true,
       );
       sink.setSubscription(subscription);
 


hoc081098:
This is the default behavior of Dart Stream. I think we should keep it

hoc081098:
The behavior you want to achieve can be implemented via an operator https://pub.dev/documentation/rxdart_ext/latest/operators/DoneOnErrorStreamExtension/doneOnError.html

rouddy:
i understand about default behavior of dart stream. but stream offers cancelOnError option.
And i think that 'resume' in  'onErrorResumeNext' means resuming next stream after previous stream pauses or stops.

but if i have alternative, i don't want to push ahead with my commit.
i will close my pr.

PR #748, opened 2024-03-10
Make web examples runnable with latest dart

dlchet:
- pubspec gets latest build_web_compilers
- README.md specifies working steps to run web examples
- minor linting issues are ignored

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -17,7 +17,7 @@ environment:
 
 dev_dependencies:
   build_runner: ^2.1.2
-  build_web_compilers: ^3.0.0
-  lints: ^1.0.1
+  build_web_compilers: ^4.0.9
+  lints: ^2.0.0
   stack_trace: ^1.10.0
   test: ^1.17.12
--- a/README.md
+++ b/README.md
@@ -226,10 +226,11 @@ Web and command-line examples can be found in the `example` folder.
  
 In order to run the web examples, please follow these steps:
 
-  1. Clone this repo and enter the directory
-  2. Run `dart pub get`
-  3. Run `dart run build_runner serve example`
-  4. Navigate to [http://localhost:8080/web/index.html](http://localhost:8080/web/index.html) in your browser
+  1. Run `dart pub global activate webdev`
+  2. Clone this repo and enter the directory
+  3. Run `dart pub get`
+  4. Run `webdev serve example`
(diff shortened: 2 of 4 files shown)


pubspec.yaml
```diff
@@ -17,7 +17,7 @@ environment:
 
 dev_dependencies:
   build_runner: ^2.1.2
-  build_web_compilers: ^3.0.0
-  lints: ^1.0.1
+  build_web_compilers: ^4.0.9
+  lints: ^2.0.0
```
hoc081098, line 21: We should keep the old version of 2 deps, because we support dart >= 2.12

dlchet: got it. i'll see if i can rework this so the example can be run on 2.12 as well as latest.

dlchet: well, i'm not sure how to make the examples work with past versions. i'm on dart 3.3.1 and i run into issues both with the old style of building
```
% dart run build_runner serve example
[SEVERE] build_web_compilers:sdk_js on asset:build_web_compilers/$package$ (cached):
Error compiling dartdevc module:build_web_compilers|lib/src/dev_compiler/dart_sdk.sound.js
Unhandled exception:
PathNotFoundException: Cannot open file, path = '/opt/homebrew/Caskroom/flutter/3.3.3/flutter/bin/cache/dart-sdk/lib/_internal/ddc_platform_sound.dill' (OS Error: No such file or directory, errno = 2)
```
and the new style
```
% webdev serve example               
[SEVERE] Pubspec errors: [The `build_web_compilers` version – 3.2.7 – is not within the allowed constraint – ^4.0.4.
null]
webdev could not run for this project.
The `build_web_compilers` version – 3.2.7 – is not within the allowed constraint – ^4.0.4.
```

So I'm a bit perplexed on what if anything should be changed. It seems to me that this problem would surface when using a recent Dart even in a separate project, so maybe I will try to make a minimal reproduction in a separate project, and then submit as an issue? @hoc081098 thank you for your attention, let me know if you see anything I'm missing

hoc081098: @dlchet I think we can use a separate `pubspec.yaml` for web example (similar to `flutter/github_search`).
Could you try this approach?

hoc081098:
Thanks, but it is obsolete now, because I have converted web example to a new project (by using melos), and it can use latest version of dependencies

PR #782, opened 2025-01-28
feat(rxdart_flutter): add ValueStream widgets

### LINKED ISSUE #779: Add ValueStreamListener and ValueStreamConsumer widgets
I would like to propose adding two new widgets to complement the existing ValueStreamBuilder:

1. **ValueStreamListener**
A widget that listens to a ValueStream and executes a callback when the stream emits new values, without rebuilding the widget tree. Useful for side effects like showing snackbars or navigation.

Signature:

```dart
ValueStreamListener<T>({
  required ValueStream<T> stream,
  required void Function(BuildContext context, T previous, T current) listener,
  required Widget child,
  Key? key,
})
```

2. **ValueStreamConsumer**

A widget that combines both ValueStreamBuilder and ValueStreamListener functionality, allowing both rebuilding and side effects from a single widget.

Signature:

```dart
ValueStreamConsumer<T>({
  required ValueStream<T> stream,
  required Widget Function(BuildContext context, T value) builder,
  bool Function(T previous, T current)? buildWhen,
  void Function(BuildContext context, T previous, T current)? listener,
  Key? key,
})
```

Notes: 

- The builder callback uses `value` instead of `data` as its parameter name to better align with ValueStream's value-focused nature and its `.value` getter. This makes the API more intuitive when working with ValueStreams.
- Unlike other similar implementations, the `listener` in ValueStreamConsumer is optional. This design choice allows the Consumer to be used primarily as a builder while maintaining the flexibility to add or remove listeners as needed, without requiring a complete widget swap.

___

I have implementations of these widgets that I've been using in my projects and would be happy to contribute them via PR in the next few days.

Let me know if you're interested in this contribution and if you'd like to see the PR.

Thanks!

GiancarloCante:
Add three new Flutter widgets for working with ValueStreams:
- `ValueStreamBuilder`: A widget that rebuilds UI based on `ValueStream` updates
- `ValueStreamListener`: A widget that performs side effects when `ValueStream` values change
- `ValueStreamConsumer`: A widget combining both builder and listener capabilities for `ValueStream`s

Features:
- Optimized for ValueStreams with required initial values
- Proper error handling for streams without values/with errors
- Optional buildWhen condition for controlling rebuilds
- Access to previous/current values in listeners
- 100% test coverage
- Detailed documentation and examples

Closes #779


hoc081098:
Thank you for your PR, I will take a look at it ASAP.

packages/rxdart_flutter/lib/src/errors.dart
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// @pragma('vm:notify-debugger-on-exception')
+ErrorAndStackTrace? validateValueStreamInitialValue<T>(ValueStream<T> stream) {
+  ErrorAndStackTrace? error;
+
+  if (!stream.hasValue) {
+    if (stream.hasError) {
+      error = ErrorAndStackTrace(
+          UnhandledStreamError(stream.error), StackTrace.current);
```
hoc081098, line 91:
```suggestion
          UnhandledStreamError(stream.error), stream.stackTrace ?? StackTrace.current);
```

packages/rxdart_flutter/lib/src/errors.dart
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+      error = ErrorAndStackTrace(
+          ValueStreamHasNoValueError(stream), StackTrace.current);
```
hoc081098, line 94:
```suggestion
          ValueStreamHasNoValueError(stream), stream.stackTrace ?? StackTrace.current);
```

packages/rxdart_flutter/lib/src/value_stream_builder.dart
```diff
@@ -1,41 +1,49 @@
-import 'dart:async';
-
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
-import 'package:rxdart/rxdart.dart';
 
-/// Signature for the `builder` function which takes the `BuildContext` and the current `data`
+import '../rxdart_flutter.dart';
```
hoc081098, line 4: I think importing 'value_stream_listener' could be better

packages/rxdart_flutter/lib/src/value_stream_builder.dart
```diff
@@ -92,182 +109,49 @@ class ValueStreamBuilder<T> extends StatefulWidget {
...
+            _currentValue = current;
+          });
         }
       },
-      onError: (Object e, StackTrace s) {
-        error = ErrorAndStackTrace(UnhandledStreamError(e), s);
-        reportError();
-        setState(emptyFn);
-      },
+      child: widget.builder(context, _currentValue!),
```
hoc081098, line 154: Is `!` required here?

packages/rxdart_flutter/lib/src/value_stream_listener.dart
```diff
@@ -0,0 +1,158 @@
...
+    if (_error != null) {
+      return;
+    }
+
+    if (!_initialized) {
+      _currentValue = stream.value;
+    } else {
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        _notifyListener(stream.value);
+      });
```
hoc081098, line 124: I don't understand why addPostFrameCallback is used here

packages/rxdart_flutter/lib/src/value_stream_consumer.dart
```diff
@@ -0,0 +1,158 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _error = validateValueStreamInitialValue(widget.stream);
+    if (_error != null) {
+      return;
+    }
+    _currentValue = widget.stream.value;
+  }
+
```
hoc081098, line 138: Should we handle didUpdateWidget?

packages/rxdart_flutter/lib/src/value_stream_builder.dart
GiancarloCante: No, T can be nullable.  
I fixed it and updated all the tests to include nullable cases.

packages/rxdart_flutter/lib/src/value_stream_listener.dart
GiancarloCante: Yes, for that implementation, but I already refactored it to avoid it and make it simpler.  

The idea is to only skip the first event when the widget is created; after that, notify everything, including the first event of the new stream if it changes.

```dart
final skipCount = _initialized ? 0 : 1;  

_subscription = stream.skip(skipCount).listen(
```

---

**Why `addPostFrameCallback`?**  

For that implementation, `addPostFrameCallback` is used because when the stream changes (other instance), the listener is notified immediately with the current value. This can be called during the build phase, which could trigger `setState()` or other UI updates that aren't allowed during build. This is what `StreamValueBuilder` does and causes the error if you don't use `addPostFrameCallback`.

packages/rxdart_flutter/lib/src/value_stream_consumer.dart
GiancarloCante: No, the value of the original stream is only used for the value of the first build, after that, it depends on the value of the listener, which already implements the didUpdateWidget. (It is verified in the tests)

packages/rxdart_flutter/lib/src/value_stream_listener.dart
hoc081098: ValueStream is a stream with a current value but does not guarantee replaying it to new listeners. For example, I have created [StateStream](https://pub.dev/documentation/rxdart_ext/0.3.0/rxdart_ext/StateStream-class.html).

So, I think skipping the current value might not be ideal in all cases. Additionally, replaying the value in BehaviorSubject and ValueConnectableStream is not part of ValueStream's contract.

I’m considering not skipping the value in initState and didUpdateWidget, or allowing users to configure this behavior. What do you think?

GiancarloCante: @hoc081098 

I hadn't considered those types of value streams.

I would think that the default behavior should apply to `BehaviorSubjects`, since they are more popular.

I already checked the `StateSubject`, and I wrote the same tests, but for something similar to a `NoReplayValueStream`, and it passed all the tests if we do this:

Add this param
```dart
  /// Whether or not the [stream] emits the last value
  /// like [BehaviorSubject] does.
  ///
  /// Defaults to `true`.
  final bool isReplayValueStream;
  ```
  
  ```dart
    final int skipCount;

    if (widget.isReplayValueStream) {
      skipCount = _initialized ? 0 : 1;
    } else {
      skipCount = 0;
      if (_initialized) {
        _currentValue = stream.value;
        WidgetsBinding.instance.addPostFrameCallback((_) {
          _notifyListener(_currentValue);
        });
      }
    }

    _subscription = stream.skip(skipCount).listen(
```

The idea is to only skip the latest value when the widget is created; after that, notify everything, including the latest value of the new stream if it changes.

what do you think?

GiancarloCante: @hoc081098 I think we should also add the optional child parameter in the builder and consumer, similar to [ValueListenableBuilder](https://api.flutter.dev/flutter/widgets/ValueListenableBuilder/ValueListenableBuilder.html)

GiancarloCante: @hoc081098 I already added support for `NotReplayValueStream` and added tests.

For tests I use `ValueSubject` from `rxdart_ext` as dev_dependency.
And added the child parameter.

I implemented everything at once because if there is any change, the tests will continue to work or will change very little.

What do you think?

hoc081098: Thank you for your update. I think it is good for now.

hoc081098:
It looks good to me.
Could you update flutter_search sample as well. It seems that the builder signature was changed (child param was added).
```
  error • The argument type 'Scaffold Function(BuildContext, SearchState)' can't be assigned to the parameter type 'ValueStreamWidgetBuilder<SearchState>'.  • lib/search_screen.dart:50:16 • argument_type_not_assignable
```

packages/rxdart_flutter/README.md
```diff
@@ -18,57 +18,143 @@
...
+        Text(
+          'Counter: $value',
+          style: Theme.of(context).textTheme.headlineMedium,
+        ),
+        if (child != null) child, // Use the stable child widget if provided
+      ],
+    );
   },
-);
+  child: const Text('This widget remains stable'), // Optional stable child widget
```
hoc081098, line 64: 👍 nice

packages/rxdart_flutter/lib/src/value_stream_listener.dart
```diff
@@ -0,0 +1,174 @@
...
+    } else {
+      skipCount = 0;
+      if (_initialized) {
+        WidgetsBinding.instance.addPostFrameCallback((_) {
+          _notifyListener(stream.value);
+        });
+      }
+    }
+
+    _subscription = stream.skip(skipCount).listen(
```
hoc081098, line 143: if skipCount <= 0, don't need to create another Stream 🙏 . 
```suggestion
    _subscription = (skipCount > 0 ? stream.skip(skipCount) : stream).listen(
```

GiancarloCante:
> It looks good to me 🙏
> 
> Could you please update the flutter_search sample as well? It seems that the builder signature has been changed (with the addition of child param).
> 
> ```
>   error • The argument type 'Scaffold Function(BuildContext, SearchState)' can't be assigned to the parameter type 'ValueStreamWidgetBuilder<SearchState>'.  • lib/search_screen.dart:50:16 • argument_type_not_assignable
> ```

I did it and added support for macOS. It had some issues when running it.

hoc081098 (approved):
Thanks for your hard work 👍 ! The CI failed, but I'll fix it in another PR 🙏.

GiancarloCante:
Hi @hoc081098, can you please publish a new version in Pub? Thanks!

hoc081098:
> Hi @hoc081098, can you please publish a new version in Pub? Thanks!

Yes, I will.

hoc081098:
https://pub.dev/packages/rxdart_flutter/versions/0.0.1 released 🙏

PR #788, opened 2025-02-24
Corrects docs and links

AlexV525:
https://docs.flutter.dev/packages-and-plugins/favorites

### DIFF
--- a/packages/rxdart_flutter/pubspec.yaml
+++ b/packages/rxdart_flutter/pubspec.yaml
@@ -1,7 +1,7 @@
 name: rxdart_flutter
 description: RxDart Flutter - Flutter Widgets that make it easy to use Streams with Flutter.
 version: 0.0.1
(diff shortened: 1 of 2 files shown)


packages/rxdart_flutter/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: rxdart_flutter
 description: RxDart Flutter - Flutter Widgets that make it easy to use Streams with Flutter.
 version: 0.0.1
-repository: https://github.com/ReactiveX/rxdart
+repository: https://github.com/ReactiveX/rxdart/tree/master/packages/rxdart_flutter
```
hoc081098, line 4: I think we should keep the old one

AlexV525: You can specify `homepage` to the same destination, but the repository is usually targeting the exact path of the package. See https://pub.dev/packages/image_picker

